PR #233, opened 2026-07-10
fix(tool): scope MCP stdio subprocess environment to an allowlist

### LINKED ISSUE #219: P1-6: MCP stdio subprocesses inherit the full secret-bearing parent environment
## Finding

**A05 Security Misconfiguration · High.** `registerStdio` sets `cmd.Env = os.Environ()` and only appends tool-specific vars — `internal/tool/manager.go:273-276`.

In the documented K8s/systemd deployments the parent env holds secrets: `DENKEEPER_LLM_*_API_KEY`, `DENKEEPER_TELEGRAM_TOKEN`, `DENKEEPER_OIDC_CLIENT_SECRET`, `DENKEEPER_API_AUTH_SESSION_SECRET`, etc. The `forbiddenEnvPatterns` denylist (`internal/tool/envresolve.go:17-22`) only blocks those names from `${...}` placeholder **resolution** — it does **not** scrub the inherited env. Any MCP tool (an arbitrary `#nosec G204` command) can call `os.Environ()` and read every secret. The Docker/Kubernetes sandbox paths are safer (they build explicit env from `opts.Env` only).

## Proposed fix

Replace `os.Environ()` with a minimal, explicit env for stdio subprocesses:

1. **Allowlist** of non-secret system vars MCP servers legitimately need (PATH, HOME, TMPDIR, LANG/LC_*, TZ, TERM, node/python runtime vars, Windows equivalents).
2. **Hard exclusion filter** applied to everything: drop any name matching `DENKEEPER_*` or the existing `forbiddenEnvPatterns` (reuse/extend `isForbiddenEnvVar`), so a secret can never reach the child even if allowlisted by accident.
3. Append the tool's own `cfg.Env` (explicit, already placeholder-resolved) last.
4. **Config escape hatch** for tools that need an extra var: `EnvPassthrough []string` on `[mcp]` (global) and/or per-tool `[tools.*]`. Passthrough names still run through the exclusion filter.

## Tests
- `buildStdioEnv` includes PATH/HOME + `cfg.Env`.
- Excludes `DENKEEPER_*` secrets even when present in the parent env (`t.Setenv`).
- A `DENKEEPER_*` name supplied via `env_passthrough` is filtered out (with warning).
- Run with `-race`.

## Context
Deferred from PR #218 (which fixed the paired P0-1 stored-XSS finding). Only one spawn site is affected: `internal/tool/manager.go:273`. Full plan lives in the PR #218 discussion / session notes.

Temikus:
## Security issue (#219, P1)

`registerStdio` in `internal/tool/manager.go` set `cmd.Env = os.Environ()` and appended the tool's `env`. In K8s/systemd deployments the parent environment holds every secret — `DENKEEPER_LLM_*_API_KEY`, `DENKEEPER_TELEGRAM_TOKEN`, `DENKEEPER_OIDC_CLIENT_SECRET`, `DENKEEPER_API_AUTH_SESSION_SECRET`, etc. Any stdio MCP tool subprocess could read all of them via `os.Environ()`. The existing `forbiddenEnvPatterns` denylist only guarded `${...}` placeholder resolution, not the inherited environment.

## Fix

New `buildStdioEnv` helper (`internal/tool/env.go`) builds a minimal explicit environment instead of inheriting everything:

- **Allowlist** of non-secret vars forwarded from the parent: `PATH`, `HOME`, `TMPDIR`, `USER`, `LOGNAME`, `SHELL`, `LANG`, `LC_*` (prefix), `TZ`, `TERM`; runtime vars `NODE_PATH`, `NODE_OPTIONS`, `PYTHONPATH`, `VIRTUAL_ENV`, `GOPATH`; Windows equivalents `SYSTEMROOT`, `COMSPEC`, `PATHEXT`, `TEMP`, `TMP`, `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, `PROGRAMFILES`.
- **Hard exclusion filter** (`isExcludedEnvVar`) applied to every forwarded name — allowlist hits AND passthrough entries — dropping any `DENKEEPER_*` name or `forbiddenEnvPatterns` match. A secret can never reach the child even if allowlisted or passed through by accident; a passthrough attempt on a protected var is logged and dropped.
- The tool's own `env` (explicit, already placeholder-resolved) is appended last so it wins over any inherited value.

**Config escape hatch**: `env_passthrough = ["VAR", ...]` on `[mcp]` (global `MCPConfig`) and per-tool `[tools.*]` (`ToolConfig`). Both merge and run through the exclusion filter. The global list is threaded from `m.mcpCfg` where the Manager already holds its MCP config.

Only one spawn site was affected (`registerStdio`); `os` is no longer imported by `manager.go`.

## Test coverage (`internal/tool/env_test.go`, run with `-race`)

- `buildStdioEnv` forwards `PATH`/`HOME` and the tool's `cfg.Env` values.
- Excludes `DENKEEPER_*` secrets present in the parent env (`t.Setenv`).
- A `DENKEEPER_*` name supplied via `env_passthrough` is still filtered out.
- Non-secret var only reaches the child when listed in `env_passthrough`.
- `LC_*` prefix matching works.
- `cfg.Env` overrides an allowlisted parent var (last-wins).

`just hook` (fmt-check + vet + lint + lint-ui + test + test-ui) passes. Docs updated in CLAUDE.md and `website/static/llms-full.txt` (the TOML-key drift test enforces the latter).

Fixes #219

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -122,6 +122,8 @@ cached_input = 0.5 # per million cached input tokens (0 = same as input)
 
 **Security**: SSRF protection, header injection prevention, env var denylist, URL/arg redaction in API responses.
 
+**Stdio subprocess env scoping** (`internal/tool/env.go`, `buildStdioEnv`): stdio MCP servers do NOT inherit the full parent environment — that would leak every `DENKEEPER_*` secret (API keys, tokens, session/OIDC secrets) into any tool subprocess. Instead the child gets a minimal explicit env: a built-in non-secret allowlist (`PATH`, `HOME`, `TMPDIR`, `USER`, `LOGNAME`, `SHELL`, `LANG`, `LC_*` prefix, `TZ`, `TERM`; runtime vars `NODE_PATH`/`NODE_OPTIONS`/`PYTHONPATH`/`VIRTUAL_ENV`/`GOPATH`; Windows equivalents `SYSTEMROOT`/`COMSPEC`/`PATHEXT`/`TEMP`/`TMP`/`USERPROFILE`/`APPDATA`/`LOCALAPPDATA`/`PROGRAMFILES`) plus the tool's own `env` (appended last, wins). Escape hatch: `env_passthrough = ["VAR", ...]` on `[mcp]` (global) and/or `[tools.*]` (per-tool) forwards extra parent vars. A hard exclusion filter (`isExcludedEnvVar`: any `DENKEEPER_*` name or the `forbiddenEnvPatterns` denylist) is applied to every forwarded name — allowlist hits AND passthrough — so a secret can never reach the child even if allowlisted or passed through by accident (a passthrough attempt on a protected var is logged and dropped). Only spawn site affected: `registerStdio`.
+
 ## External REST API
 
 `internal/api/` — HTTP API server (enabled by default). Auth via Bearer token (API keys) or session cookies (password/OIDC).
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -669,6 +669,11 @@ type MCPConfig struct {
 	// URLAllowlist restricts which hosts SSE tool servers may connect to.
 	// Supports wildcards (e.g. "*.internal.corp"). Empty = all non-blocked hosts allowed.
 	URLAllowlist []string `toml:"url_allowlist"`
+	// EnvPassthrough names additional parent-process environment variables to
+	// forward into every stdio MCP subprocess, on top of the built-in non-secret
+	// allowlist. Secret-bearing names (DENKEEPER_* and the forbidden-pattern
+	// denylist) are always filtered out even if listed here.
+	EnvPassthrough []string `toml:"env_passthrough"`
 }
 
 // ToolConfig defines an MCP tool server to spawn.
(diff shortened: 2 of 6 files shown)


Temikus:
Confirmed — the finding is valid. `isAllowlistedEnvVar` did an exact-match against the uppercase `stdioEnvAllowlist` (and the `LC_` prefix check ran on the raw name), and the `env_passthrough` set was keyed by the operator's literal spelling. Meanwhile `isExcludedEnvVar` already uppercase-normalizes (via `strings.ToUpper` + `isForbiddenEnvVar`). So on Windows, where env names are case-insensitive and surface mixed-case (`Path`, `SystemRoot`, `ComSpec`), the advertised Windows equivalents would be silently dropped even though they were in the allowlist.

Fix (`internal/tool/env.go`):
- `isAllowlistedEnvVar` now uppercases the candidate before the map lookup and the `LC_` prefix check.
- The `env_passthrough` set is keyed by uppercase, and the parent-env name is uppercased before the membership test — so case-insensitive passthrough matching too.
- The exclusion filter was already case-insensitive; left unchanged (verified a mixed-case `Denkeeper_Api_Key` is still dropped).
- Forwarded `KEY=value` entries keep the parent's **original spelling** — only the comparison is normalized (`env = append(env, kv)` uses the raw entry).

Added Windows-shaped tests in `env_test.go`: `TestBuildStdioEnv_WindowsMixedCaseNames` (mixed-case `Path`/`SystemRoot`/`ComSpec` forwarded with original spelling, non-allowlisted `windir` dropped), `TestBuildStdioEnv_MixedCaseSecretStillExcluded`, and `TestBuildStdioEnv_PassthroughCaseInsensitive`.

Local verification:

```
$ go test -race -run TestBuildStdioEnv ./internal/tool/
ok  	github.com/Temikus/denkeeper/internal/tool	1.567s
```

`just hook` is fully green (fmt-check, vet, lint, lint-ui, test, test-ui).

PR #246, opened 2026-07-25
fix(telegram): stop silent markdown corruption by rendering to HTML locally

### LINKED ISSUE #243: Telegram replies silently corrupted by legacy Markdown parse mode
Agent replies are sent with `parse_mode=Markdown` (Telegram's legacy mode) and nothing escapes the text — `internal/adapter/telegram/telegram.go:331`, same again at `:540`. That parser has no intraword-emphasis rule, so underscores inside URLs are consumed as italic delimiters.

Observed in a delivered message. Audit log had:

```
.../flash/7/NRR-JaUmg_hm7jYl-LiWx/gmail-mcp-servers-brief-v2.md
```

Telegram rendered:

```
.../flash/7/NRR-JaUmghm7jYl-LiWx/gmail-mcp-servers-brief-v2.md
```

The `_` in that URL paired with the `_` in the next URL in the same message. Everything between the two rendered italic, and a `**HTML**` caught inside that span came out as literal asterisks — legacy mode won't nest entities.

**Impact.** Telegram autolinks post-parse text, so the anchor target is missing the underscore too — the link 404s. Anything containing `_`, `*`, `` ` `` or `[` is exposed: URLs, snake_case identifiers, file paths. LLM output contains these constantly.

**Why nothing caught it.** Telegram accepted the message successfully. The plain-text retry at `:334` only fires on a send error, so it can't see a misparse. There are no metrics anywhere under `internal/adapter/`. We have no way to tell how long this has been happening or how often.

**Repro.** Any agent reply containing two URLs that each have an underscore in the path.

Remediation is scoped separately — this issue tracks the defect only.

jamestelfer:
Fixes #243

Full requirements, delivery plan and implementation notes: **https://dollop.telfer.me/keep/eminent-hawk/index.html**

## Purpose

Agent replies were reaching Telegram as raw markdown under `parse_mode=Markdown`, Telegram's legacy parser, with nothing escaping the text. That parser has no intraword-emphasis rule, so an underscore inside a URL opened an italic entity and was consumed as markup. Identifiers came out mangled and autolinked destinations stopped resolving.

The failure was silent. Telegram accepted these messages successfully, so the existing retry-on-error safety net never fired: corruption was unbounded in frequency and invisible in logs and metrics. The blast radius is any identifier containing `_`, `*`, `` ` `` or `[` — which LLM output produces routinely — and mangled URLs become dead links. Full analysis: [Problem Statement](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#problem-statement).

This stops asking Telegram to parse markdown at all. Markdown is parsed locally as CommonMark and rendered to the HTML subset Telegram supports, then sent with `parse_mode=HTML`. A compliant CommonMark parser does not treat an intraword underscore as emphasis, so the defect is removed at the root rather than escaped around. HTML parse mode also has a three-character escape surface instead of a dozen-plus, which makes the remaining failure modes tractable. Rationale, including [why not MarkdownV2](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#decision-13-in-detail-why-not-markdownv2): [Solution](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#solution).

Because the renderer returns a sequence of independently tag-balanced blocks rather than one opaque string, it also closes an adjacent failure mode that had no handling on this path: Telegram's 4096-character message cap. Long replies are now split across several messages at safe boundaries, and a single block that exceeds the cap has its wrapping elements closed and reopened across the split.

Rendering is confined to the Telegram adapter. Message text stays canonical markdown at the adapter interface, so Discord — which relies on Discord's own native markdown rendering — is untouched. No configuration surface is added: the only path a toggle could select is the defective one, and a plain-text fallback covers rendering or send failure without operator intervention.

## Context

- 35 EARS requirements across [renderer](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#renderer), [chunker](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#chunker), [adapter](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#adapter) and [observability](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#observability), each mapped to a phase in the [coverage matrix](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.plan-1.html#requirements-coverage-matrix).
- Commits map one-to-one onto the [delivery phases](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.plan-1.html#plan-telegram-markdown-rendering), so the history reads in the same order as the plan. Decisions made while implementing are recorded in [progress](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.progress-1.html).
- Manually tested end to end against a live installation, which is currently running this branch.
- The renderer is derived from `leonid-shevtsov/telegold` (MIT, attribution and licence retained) and vendored rather than imported: upstream has had one commit since 2022, pins goldmark three minor versions behind, and carries six defects measured by executing it. That exercise also corrected three requirements — see [Implementation Decisions](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#implementation-decisions) and [Verified upstream findings](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.plan-1.html#verified-upstream-findings).

## Out of scope

[Phase 7](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.plan-1.html#phase-7-fallback-counter-and-warning-log) — a counter and warning log on every plain-text fallback (requirements 32 and 33) — is not included. Fallbacks currently emit neither a metric nor a warning, and the adapter layer has no metrics at all, so this class of defect stays hard to detect. Happy to follow up if you want it; it does not affect correctness of the send path.

Also out of scope, and now tracked separately: Telegram approval prompts can exceed the message cap and silently lose the approval keyboard (Temikus/denkeeper#252). The length is decided in the dispatcher's activity log, not here — the adapter's explicit-parse-mode contract is what makes the overflow fatal rather than merely ugly, so the fix belongs where the length is set. This branch corrects only the comment that wrongly implied the two chunk budgets were equivalent.

## Notes for reviewers

- `extension.GFM` is deliberately not used: it bundles `linkify`, which would autolink the exact underscore-bearing URLs this work exists to fix. `Table` and `Strikethrough` are selected individually.
- The activity log is the highest blast-radius surface. It passes explicit `ParseMode: "HTML"` with hand-built markup, and the explicit-parse-mode pass-through contract keeps it unrendered.
- There are two send paths, `Send` and `SendAndGetID`, each previously running its own Markdown-then-retry sequence. Both are covered.
- Tables and images degrade rather than fail closed — tables to aligned preformatted blocks, images to anchors labelled with their alt text. Raw HTML is the sole fail-closed case, and its error names the construct.


jamestelfer:
@Temikus this should be ready. Sorry for the size: it could be broken up into the renderer and chunker if you were keen.

In theory the renderer could be a separate library but I didn't want to commit you to that. 

It's running quite well for me, handling markup as well as could be expected without doing things like rendering tables as images

Temikus:
@jamestelfer Thank you so much! Lemme run a quick CI check.

Temikus:
@jamestelfer Small question - why derive telegold's renderer and vendor it instead of writing the renderer against goldmark's AST?

jamestelfer:
> @jamestelfer Small question - why derive telegold's renderer and vendor it instead of writing the renderer against goldmark's AST?

The goal was to [use the prior art as a baseline](https://dollop.telfer.me/keep/eminent-hawk/telegram-markdown-rendering.prd.html#implementation-decisions) with some improvements rather than implementing it from scratch. Ideally this led to a more solid and complete implementation than would otherwise have been the case.

The model could probably have done that quite happily, but it probably would have used the MIT licensed code as an exemplar anyway at that point.

Happy to change tack -- what would you prefer?

jamestelfer:
Checked the Grype failure: this came from vulnerabilities on `main` that have now been fixed. The next run should be clear.

Temikus:
@jamestelfer I would prefer we don't vendor as there's not much to it, if that's ok.

jamestelfer:
It's different to the original as a number of bugs have been fixed. 

I can make it a fork on my account for now and see if it gets accepted upstream, or just maintain it there. What's your preference?

jamestelfer:
Created a fork in my account and stripped out the vendored copy.

The updates against the original upstream that hardened the URL handling and fixed encoding bugs, as well as widening support to tables etc, are all in the new fork. The chunked rendering made it a quite different API, so these  changes are not over-likely to be upstreamed.

What's left is far slimmer: about 300 lines of production code and a large amount of test code ensuring that we handle the chunking and failure fallbacks properly.

The CI failure from betterleaks flows in from "main", and Claude is struggling because this is a fork and it doesn't have permission to auth. 

This is now quite a long series of commits. If you would like them reworked let me know.

jamestelfer:
Changing to a draft as I rework this

jamestelfer:
Okay, this is now fully reworked and ready to go.

Temikus:
@claude please review this for project conventions

Temikus:
@jamestelfer thanks! Looking at this now. Thanks for fixing up mise too, much better now.

Temikus:
LGTM, thanks @jamestelfer ❤️

PR #247, opened 2026-07-25
chore(deps): bump golang.org/x/text to v0.39.0 for GO-2026-5970

Temikus:
## Summary

Bumps the indirect `golang.org/x/text` requirement from v0.38.0 to v0.39.0 to fix **[GO-2026-5970](https://pkg.go.dev/vuln/GO-2026-5970)** — infinite loop on invalid input in `golang.org/x/text/unicode/norm`.

This is a security fix, not a routine refresh. `govulncheck` reports the vulnerability as **reachable from our code**, via two call paths:

- `internal/api/server.go:404` — `api.Server.Run` → `http.Server.Shutdown` → `norm.Form.*`
- `internal/tool/oauth/handler.go:51` — `oauth.oauthRoundTripper.RoundTrip` → `http2` → `norm.Form.*`

## Why now

The `Security Scan` CI job runs `govulncheck ./...` and has been failing on **every open PR** because of this — including PRs that touch no Go code at all (e.g. the npm-only prettier and brace-expansion bumps). It was surfacing as a per-PR failure while actually being a pre-existing condition on `main`, which made it easy to wave through as unrelated noise. Fixing it at the source clears the job repo-wide.

## Changes

- `go.mod`: `golang.org/x/text` v0.38.0 → v0.39.0 (indirect)
- `go.sum`: `go mod tidy` additionally drops stale entries left behind by earlier dependency merges (old `goja` digests, `go-toml` 2.4.0/2.4.2, `go-oidc` 3.18.0) and records the `x/mod` v0.37.0 / `x/tools` v0.47.0 hashes that x/text v0.39.0 references. Those two are `go.sum`-only — not in `go.mod`, not in the build graph.

No source changes; no import or API surface change.

## Verification

| Check | Result |
|---|---|
| `govulncheck ./...` | **No vulnerabilities found** (was: 1 reachable) |
| `go test -race ./...` | pass, all packages |
| Web UI tests | pass — 722 tests / 41 files |
| `lint` / `vet` / `fmt-check` | clean |

## Note on remaining alerts

Dependabot still reports other vulnerabilities on the default branch. This PR addresses the one `govulncheck` flags as reachable from our call graph; the remainder are in modules we require but do not call. The `Vulnerability Scan (Grype)` job scans the whole filesystem at `--fail-on high` and may therefore still fail on those — that is pre-existing and out of scope here.

https://claude.ai/code/session_01V29ewnvKNpWUGiH4EBMygX


Temikus:
### CI status: 16/17 green — the one red is pre-existing on `main`

**`Security Scan` now passes.** That is the job this PR targets: `govulncheck ./...` went from *1 reachable vulnerability* to **`No vulnerabilities found`**. It had been failing on every open PR, including ones touching no Go code at all.

**`Vulnerability Scan (Grype)` still fails, and it is not caused by this change.** It fails on the base branch independently:

| Ref | Commit | `Vulnerability Scan (Grype)` |
|---|---|---|
| `main` | `1ce7ee48` (base of this PR) | ❌ failure — [run](https://github.com/Temikus/denkeeper/actions/runs/30148439216) |
| `main` | `29f12a13` | ❌ failure |
| `main` | `cc3d9105` | ❌ failure |
| this PR | `829d0ba` | ❌ failure (same) |

It also failed identically on #239, #240 and #242 — three PRs with unrelated diffs (a Go digest bump, an npm-only prettier bump, a grpc bump). It last passed on `main` on 2026-07-20, so it began failing when new advisories landed in the Grype DB, not from any diff.

This PR cannot have introduced it: the only change is an *upward* bump of `golang.org/x/text`, which can remove Go-side matches but never add them.

**What I could not determine:** which package Grype is actually flagging. The job log only prints `Found vulnerabilities with level 'high' or higher` — the per-package detail goes to the SARIF upload, visible under Security → Code scanning. Worth a look there, since `--fail-on high --only-fixed` means whatever it is has a fix available. Separately, GitHub reports 5 Dependabot alerts on the default branch (4 high, 1 low); this PR clears the one `govulncheck` proved reachable from our call graph, and the rest are in modules we require but do not call.

Recommendation: treat the Grype failure as a separate pre-existing issue rather than a blocker here — this PR strictly improves the security posture and turns one of the two red security jobs green.

---

Issue #252, opened 2026-07-28
Telegram approval prompts can exceed the message cap, silently losing the approval keyboard


jamestelfer:
An approval prompt sent through the Telegram activity log can exceed Telegram's 4096-character cap, in which case the message is rejected outright and the approval keyboard never arrives. The approval then waits for a click that cannot happen, until it times out. Nothing is logged above `Debug`.

Predates #246; found while reviewing it. The adapter side is correct — this is a length decided in `internal/agent/dispatcher.go`.

## Why the length escapes its budget

`approvalArgsMaxChars = 2000` (`dispatcher.go:1682`) truncates tool arguments counting **unescaped runes**. `html.EscapeString` (`:1523`) then expands `&` to `&amp;` — five bytes for one — and `<`/`>` fourfold. Measured on the rendered output:

```
args of "&" x2000 -> 10071 bytes   (Telegram cap 4096, chunk budget 3500)
args of "<" x2000 ->  8071 bytes
args of ">" x2000 ->  8071 bytes
```

The overflow check at `:1586` cannot absorb it: it rolls to a fresh chunk only when `len(c.lines) > 1`, so a chunk whose bulk *is* the approval section has nowhere to move. A variant needing no escape expansion at all reaches 4748 bytes — 200 accumulated tool lines plus a 2000-rune approval on a chunk whose `messageID` is still empty, so `setPending` at `:1685` does not roll a fresh chunk.

The adapter then forwards it unchunked and unretried, which is correct by contract: an explicit `ParseMode` means the caller owns its own markup, so the adapter must not re-render or re-split it (requirement 28). That contract is what turns a formatting problem into a functional one.

Tool arguments containing a URL query string, or JSON or XML with `&`, are routine.

## Reproduction

Trigger a supervised approval whose tool arguments contain roughly 2000 ampersands — a URL with a long query string will do. The approval message is rejected with `Bad Request: message is too long`, and no buttons appear.

## Suggested direction

Budget the **escaped** length rather than the source length: either truncate after escaping, or measure the escaped width while truncating. The check at `:1586` should also be able to split a chunk whose approval section alone exceeds the budget, rather than giving up when there is only one line.

## Companion: the failure is invisible

[...]

PR #258, opened 2026-07-29
fix(telegram): render agent replies as escaped HTML, not legacy Markdown

### LINKED ISSUE #243: Telegram replies silently corrupted by legacy Markdown parse mode

Temikus:
## Problem

Agent replies were sent with `parse_mode=Markdown` — Telegram's *legacy* parser — with nothing escaping the text (`internal/adapter/telegram/telegram.go:331`, again at `:540`).

That parser has no intraword-emphasis rule, so an underscore inside a URL path pairs with the next underscore anywhere in the message. In the reported case both `_` characters were deleted, everything between them rendered italic, and a `**HTML**` caught inside the span came out as literal asterisks. Because Telegram autolinks post-parse text, the anchor target lost the underscore too and the link 404'd.

Anything containing `_`, `*`, `` ` `` or `[` was exposed — URLs, `snake_case` identifiers, file paths — which is most of what an LLM writes.

Telegram *accepts* these messages, so the plain-text retry at `:334` (which only fires on a send error) could never observe the misparse.

## Approach

Replies with no explicit parse mode are now parsed as **CommonMark** and re-emitted as the restricted **HTML** subset Telegram accepts, with every text node escaped.

Why HTML rather than MarkdownV2 or plain text:

- **HTML is already the codebase's parse mode for anything with structure** — the dispatcher's activity log and approval prompts (`internal/agent/dispatcher.go`) build Telegram HTML with `html.EscapeString` on text nodes. This extends the pattern instead of introducing a second one.
- **Escaped-text-node HTML is the only option where correctness does not depend on getting an escape table right.** MarkdownV2 needs 18 characters escaped and still requires a markdown parser to know which occurrences are markup; a mistake there reintroduces exactly this class of bug. In HTML, text is text.
- **Plain text would work but regresses formatting** users rely on — code blocks, bold, links.
- Parsing is done with **goldmark**, which was already in the module graph (test-only dependency of `html-to-markdown/v2`, pinned in `go.sum`, no new module version). It is the same CommonMark dialect `marked` gives the web dashboard, so Telegram and the dashboard now agree on how a reply looks.

The key property: goldmark's **linkify** extension consumes a bare URL before emphasis delimiters are collected, so underscores inside URLs stay literal *by construction* rather than by relying on flanking rules. CommonMark's intraword rule covers the rest (`max_tool_rounds`, `skill_name_v2`, …).

### What changed

- **New `internal/adapter/telegram/render.go`** — markdown → Telegram HTML. Emits only tags Telegram allows (`b i u s a code pre blockquote`). Headings become bold, lists become `•`/numbered lines, tables stay literal pipe text, raw HTML and unsupported link schemes (`javascript:`, relative paths) degrade to escaped text so nothing is dropped.
- **`prepareOutgoingText` / `sendWithFallback`** centralise the parse-mode decision for both `Send` and `SendAndGetID`.
- **Scope guard:** a caller-supplied `ParseMode` means the caller owns its markup — the dispatcher's activity log and standalone approval prompts — and is passed through untouched **and never retried**. `EditText`/`EditMessage` are unchanged.
- **No new failure mode:** HTML escaping lengthens text, so a render crossing the 4096-character cap falls back to the unmodified source with no parse mode. A render Telegram rejects for any other reason falls back the same way.

Rendered example:

```
Report is at <a href="https://example.com/flash/7/NRR-JaUmg_hm7jYl-LiWx/brief-v2.md">https://…/NRR-JaUmg_hm7jYl-LiWx/brief-v2.md</a> and the <b>HTML</b> copy is at …
```

## Tests

`internal/adapter/telegram/render_test.go` (24 cases) and `send_test.go` (11 cases):

- **The reported repro** — two URLs each with an underscore in the path: both paths intact, no `<i>` emitted, `**HTML**` between them still bold, and both `href` targets carry the exact path.
- **Round-trip invariant** — prose containing `max_tool_rounds`, `skill_name_v2`, `3 * 4 * 5`, `a[0] < b[1] && c > d`, `` `code_with_underscores` ``, `[bracket`, a trailing `_underscore` and a query-string URL: stripping tags and decoding entities from the render returns the source **character for character**.
- **Wire level** — a stub Bot API transport captures the actual request body and asserts `parse_mode=HTML`, that the delivered `text` round-trips, and that a rejected send retries once with the unmodified source and no parse mode. Caller-owned markup is verified not to be rewritten or retried.
- Formatting not regressed: bold, italic, strikethrough, inline code, fenced blocks (with and without a language), links, headings, lists, nested lists, blockquotes.
- Safety: raw `<script>` is escaped and renders back as literal text; `javascript:` links render as label only; an unusable code-fence info string is dropped; no tag outside Telegram's allowlist is ever emitted; concurrent renders are `-race` clean.

`just hook` passes (fmt-check, vet, lint, lint-ui, test, test-ui). `just test-integration` passes. `just scan` reports only the three pre-existing `//nolint`-annotated G118 findings.

## Docs

- `website/content/docs/guides/telegram-setup.md` — new "Message formatting" section: what markdown maps to, the escaping guarantee, and the plain-text fallback.
- `CLAUDE.md` — invariant describing the empty-vs-explicit `ParseMode` contract and the tag allowlist constraint on the renderer.

Closes #243

https://claude.ai/code/session_01JsdhCwSR1uZ5dcYziVBvWp


Temikus:
Closing as a duplicate of #246, which fixes the same defect (#243) and predates this branch by several days.

#246 is the better implementation and should be preferred:

- It handles Telegram's 4096-character cap for long replies, splitting at safe boundaries and reopening wrapping elements across the split. This branch only falls back to unmodified source text on overflow.
- Its renderer is a dedicated, separately-tested package with table and image degradation paths, derived from `telegold` with upstream defects corrected.
- It has been manually tested end to end against a live installation.

One correction worth recording, since this branch argued the opposite: I claimed goldmark's `linkify` extension was load-bearing for the fix — that underscores in URLs survive because linkify consumes the URL before emphasis delimiters are collected. That is wrong. Measured on goldmark directly:

```
input:  See https://x.com/flash/7/NRR-JaUmg_hm7jYl-LiWx/a.md and https://y.com/p/b_c_d/e.md plus **HTML** here.

no linkify:   ...NRR-JaUmg_hm7jYl-LiWx/a.md ... b_c_d/e.md ... <strong>HTML</strong>
with linkify: <a href="...NRR-JaUmg_hm7jYl-LiWx/a.md">...</a> ... <strong>HTML</strong>
```

Both preserve every underscore. CommonMark's intraword-emphasis rule already does the work, so linkify is neither necessary nor harmful here — #246's decision to omit it is safe, and its regression test pins real behaviour.

No functional coverage is lost by closing this. Issue #243 stays assigned to #246.

The approval-prompt overflow that #246 scoped out (#252) is handled separately in #257.

PR #283, opened 2026-08-06
fix(oauth): persist rotated refresh tokens on silent auto-refresh

jamestelfer:
## Purpose

Fixes "refresh token reuse detected" errors from the Todoist MCP tool, which force a full re-auth. Any provider with refresh-token rotation hits this on every process restart, tool restart, health-check auto-restart, or reload after the first silent refresh.

Verified against Todoist: the refresh token is updated back to the persistent store when it changes, and the connection persists.

## Context

`Handler`'s cached token source refreshed tokens in memory only — the rotated refresh token was never written back to the store. Since a `Handler` is recreated fresh on every reconnect, it reloads the stale, already-consumed refresh token and replays it. Providers that rotate refresh tokens correctly reject that as reuse and revoke the token family. The provider's behaviour is correct; the bug is ours.

`golang.org/x/oauth2` has no refresh-notification hook (confirmed via docs) — `TokenSource` is a single-method interface, and `ReuseTokenSource*` only caches the last token in memory. Fix: a small notifying wrapper inside the reuse-caching layer that persists on every real refresh.

### DIFF
--- a/internal/tool/oauth/handler.go
+++ b/internal/tool/oauth/handler.go
@@ -15,22 +15,15 @@ import (
 	"golang.org/x/oauth2"
 )
 
-// Provider compatibility notes:
+// Providers vary: some issue non-expiring tokens with no refresh token
+// (no refresh needed); others (e.g. Todoist) issue expiring tokens with a
+// rotating refresh token, which must be persisted on every refresh or a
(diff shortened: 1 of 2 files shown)


Temikus:
@jamestelfer thank you! 🙏

PR #343, opened 2026-08-19
feat(eval): suggest test cases from history (Stage D3 backend)

Temikus:
Stage D3 backend: the telemetry query and endpoint behind "Suggest test cases from history". No UI in this PR.

The Evals page has no in-page fill path - a new user lands on an empty task set whose only sources live on other screens. This is the cold-start fix.

## What changed

**`agent.ListInterestingTurns(ctx, agent, since, limit)`** pairs each user message with the assistant reply that answered it (next assistant row in the conversation) and rolls up that reply's `tool_calls` - count, max round, rejected/failed - plus the turn's `message_skills` command matches, scoped by `conversation_stats.agent`. It also carries the last 4 preceding messages for pinned history, each capped at 2000 runes.

It sits on a new narrow `InterestingTurnStore` interface, type-asserted from `MemoryStore` the way the API layer already reaches `TelemetryStore` - rather than widening `MemoryStore` and every hand-written mock.

**`eval.Suggest`** scores each turn by signal count and stratifies:

- Signals: `tool_fault` (any rejected/failed outcome), `many_rounds` (max round >= 3), `high_cost` (reply cost in the pool's top decile), `command_skill`.
- Category: command match -> `skill_command`; content opening with the `[Scheduled` header -> `scheduled`; >= 3 tool calls or rounds -> `tool_heavy`; else `chat`.
- **Stratified, never top-N overall**: `limit/4` per category, spare slots redistributed round-robin. A set drawn purely by interestingness is all failures and represents nothing the agent normally does.
- A turn with zero signals is never offered, and `Store.SavedTaskSources` subtracts turns already saved as a task (`source_conversation_id` + `source_message_id`), so an accepted suggestion does not resurface.

**`GET /api/v1/eval/suggest?agent=&limit=&since=`** (scope `eval:read`, default limit 20 / max 100, default window 90 days). It writes nothing - accepting a candidate is a separate call to the task create endpoint. `llms.txt` lists it; spec regenerated.

Response shape (the D3 UI slice builds cards against this):

```json
{
  "candidates": [
    {
      "prompt": "dig through the logs",
      "category": "tool_heavy",
      "conversation_id": "chan:main",
      "message_id": 4211,
      "created_at": "2026-08-14T09:12:33Z",
      "signals": ["tool_fault", "many_rounds"],
      "preceding": [{ "role": "user", "content": "..." }]
    }
  ]
}
```

## Notes for the reviewer

- The `since` bound is formatted to SQLite's own `DATETIME` text rather than bound as a `time.Time`. `created_at` is written by `CURRENT_TIMESTAMP` in that layout and the comparison is a string comparison whichever way the driver renders the parameter - binding a `time.Time` returned zero rows.
- Preceding context is fetched per candidate, so the pool size is also the query count. Hence the 500 clamp in the store and the `limit * 10` (min 100) pool in the handler.
- Only `match_type = 'command'` counts toward the command signal; ambient and scheduled matches deliberately do not.

## Test evidence

`just hook` green (fmt-check, vet, lint, lint-ui, test, test-ui, openapi-check), before and after `git rebase origin/main`.

New tests, all `-race`:

- `internal/agent/memory_interesting_test.go` - reply telemetry roll-up (cached calls excluded from faults), ambient/scheduled matches ignored, agent scoping, preceding context oldest-first, the content cap, unanswered turns skipped, the `since` bound.
- `internal/eval/suggest_test.go` - zero-signal drop, all four categories, command beats tool weight, stratification against 8 failures + one of each other kind, leftover-slot redistribution, score-then-recency ordering, top-decile relative to the pool, exclusion, and `SavedTaskSources`.
- `internal/api/eval_suggest_test.go` - seeded against the real in-memory SQLite store (no canned candidates): one interesting turn per category plus a boring one returns exactly four, one per category; an already-saved turn is excluded; agent scoping; default and explicit limit; malformed params 400; scope gating (403 without `eval:read`, 200 with it); 503 unwired.

### DIFF
--- a/internal/agent/memory.go
+++ b/internal/agent/memory.go
@@ -199,9 +199,10 @@ type SQLiteMemoryStore struct {
 
 // Compile-time interface checks.
 var (
-	_ MemoryStore        = (*SQLiteMemoryStore)(nil)
-	_ TelemetryStore     = (*SQLiteMemoryStore)(nil)
-	_ ActiveChannelStore = (*SQLiteMemoryStore)(nil)
+	_ MemoryStore          = (*SQLiteMemoryStore)(nil)
+	_ TelemetryStore       = (*SQLiteMemoryStore)(nil)
+	_ ActiveChannelStore   = (*SQLiteMemoryStore)(nil)
+	_ InterestingTurnStore = (*SQLiteMemoryStore)(nil)
 )
 
 const schema = `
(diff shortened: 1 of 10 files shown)


Temikus:
@claude rebase this onto master please - we fixed some CI issues.

PR #354, opened 2026-08-25
docs(eval): document the eval loop beyond curl

Temikus:
Stage D slice D4 (docs half) of `design/eval-subsystem.md` §8. Docs and generated artefacts only - no feature code.

## What changed

1. **README "Evals" section** - was written when a run was API-only. Now covers the shipped surface: the three fill paths (Save as test case, `GET /eval/suggest`, JSONL import/export), `sample_tasks` and `POST /eval/estimate`, live-engine execution with its two bounds, the scorecard, and the judging flow via `/judge-eval`.
2. **New `website/content/docs/concepts/evals.md`** (slug `evals`, weight 57, between Dry Runs 55 and Observability 60) - the loop end to end, test sets, runs/bounds/isolation, judging and blinding, the verdict rule, dashboard state, config. Linked from `concepts/dry-run.md`, `guides/mcp-server.md`, the `guides/web-dashboard.md` page table, and the concepts index.
3. **Drift fixes** - the website REST reference was missing `GET /eval/suggest`, `GET /eval/runs/{id}/pairs`, and the `sample_tasks`/`as_of` run inputs; `.claude/rules/rest-api.md`'s endpoint map had no eval rows at all.

The web page is described only as far as it has actually shipped (#345). The results view and suggest-from-history cards - in flight on other branches - are called out in a callout with the endpoints to use meanwhile, rather than documented ahead of the code.

## Notes for the reviewer

- **`internal/api/llmstxt.go` needed no change.** The D4 plan expected `estimate`/`suggest`/`pairs` to be missing from it; all eight eval entries are already there and their scopes match the routes in `server.go`. Checked, not assumed.
- **`just openapi` produced a zero diff** - the committed spec was already current. `just openapi-check` passes.
- Every claim was verified against the code: limit default 20 / max 100 and the 90-day `since` window from `internal/api/eval.go`, the three estimate bases from `internal/eval/estimate.go`, the win/loss/tie/pending outcomes from `pairview.go`, the nine `[eval]` defaults from `internal/config/config.go`.
- All 28 internal `/docs/...` links in the tree resolve against front-matter slugs (checked programmatically). No fragment anchors used, so nothing new can break the way #346 fixed.

`just hook` green: fmt-check, vet, lint, lint-ui, test, test-ui, openapi-check.

https://claude.ai/code/session_01UtYGcxeWoLiy3EcQrvNfTf


Temikus:
Post-merge review — two fact-check findings; everything else in the new docs checked out against the actual routes, scopes, body fields, verdict rule and config keys:

1. **The judge-key security claim overstates what the scope prevents** (`website/content/docs/concepts/evals.md`, Judging section). *"Give it a key scoped to `eval:read,eval:write` and nothing more: that key can judge a run but cannot start one"* — the **MCP tool surface** can't start runs (no run-creation tool exists among `eval_*`, by design), but the key itself can: `POST /api/v1/eval/runs` requires exactly `eval:write` (`internal/api/server.go:285`), and the same Bearer key authenticates against the REST API on the same server — this PR's own REST reference documents that scope. Worth rewording to attribute the "cannot start" guarantee to the MCP tool surface rather than the key, since as stated the "right shape given runs spend real tokens" rationale doesn't hold.

2. `.claude/rules/rest-api.md:36` has an unescaped pipe inside a code span — `` `{name}/export|import` `` — which still delimits cells in a GFM table, splitting the row and shifting the scope/semantics columns right. The very next line escapes correctly (`summary\|samples\|pairs`), as does the pre-existing tools row.

---

PR #355, opened 2026-08-25
feat(eval): suggest test cases from history on the Evals page

Temikus:
Client half of Stage D slice D3. The `GET /eval/suggest` backend shipped in #343; this is the surface that uses it.

## What it does

New `web/src/components/SuggestCases.svelte` renders the stratified candidates as cards under the launcher: prompt, category chip, and a "why" line translating the API's signal names into operator English (`tool_fault` -> "a tool call was rejected or failed").

- **Accept** adds the case to the chosen test set via `addEvalTask`, creating the set on the fly and pinning the candidate's preceding turns as history, the way `SaveTestCase.svelte` does.
- **Reject** hides the candidate for the session only. Nothing is written, so a refresh offers it again - the copy says so.
- **Batch accept** via per-card checkboxes, with a progress label and a partial-failure message naming how many landed.

"Suggest from history" becomes the empty state's first CTA, completing the three the design specifies, plus a header toggle once a test set or run exists. The suggest and import panels take turns so they cannot stack above the launcher.

## Review notes

- The diff to `Evals.svelte` is deliberately confined to the mount points (46 lines: import, toggle state, two triggers, the collapsible block, and an `onaccepted` callback that re-reads the test sets). The results-panel region is untouched - a parallel D2 branch is editing it.
- `web/src/test/handlers.js` had a `/eval/suggest` handler returning a bare `[]`; the endpoint actually answers `{candidates: [...]}`. Fixed, with a fixture carrying one candidate per category.
- `Evals.test.js` carried an assertion that "Suggest" was absent from the empty state, per the D1 note "until it lands the CTA is absent". It has landed, so that assertion is inverted.
- A `ui-reviewer` pass ran over the component; its findings are applied. The substantive one: success feedback lived inside the `visible.length > 0` branch, so "select all, accept" emptied the list and swallowed its own confirmation.

<details>
<summary>Checks</summary>

`just test-ui`: 931 passed (19 new). `just hook`: fmt-check, vet, lint, lint-ui, test, test-ui, openapi-check all green.

Note for the other eval worktrees: `just lint` first failed citing files under `../feat-eval-d2-results/`. The golangci-lint cache under `~/Library/Caches/golangci-lint` is shared across worktrees; a cache clean cleared it and the same tree linted 0 issues.
</details>

https://claude.ai/code/session_013KnwwPWFVDU4UQQ8jW9zzg

### DIFF
--- a/web/src/__tests__/pages/Evals.test.js
+++ b/web/src/__tests__/pages/Evals.test.js
@@ -31,7 +31,7 @@ afterEach(() => {
 })
 
 describe('Evals page — empty state', () => {
-  test('teaches the loop and offers Chat and import, but not suggest', async () => {
+  test('teaches the loop and offers all three CTAs', async () => {
     emptyInstance()
     render(Evals)
 
@@ -39,10 +39,21 @@ describe('Evals page — empty state', () => {
     expect(
       screen.getByText(/Save real conversations as test cases, then compare your current model against a candidate on them/)
     ).toBeInTheDocument()
+    expect(screen.getByTestId('empty-suggest-cta')).toBeInTheDocument()
     expect(screen.getByTestId('empty-chat-cta')).toBeInTheDocument()
     expect(screen.getByTestId('empty-import-cta')).toBeInTheDocument()
-    // "Suggest from history" ships in a later slice — absent, not disabled.
-    expect(screen.queryByText(/Suggest/i)).not.toBeInTheDocument()
+  })
+
+  test('the suggest CTA opens the candidate cards', async () => {
+    emptyInstance()
+    render(Evals)
+
+    await waitFor(() => expect(screen.getByTestId('evals-empty')).toBeInTheDocument())
+    expect(screen.queryByTestId('suggest-panel')).not.toBeInTheDocument()
+
+    await fireEvent.click(screen.getByTestId('empty-suggest-cta'))
+    await waitFor(() => expect(screen.getByTestId('suggest-cards')).toBeInTheDocument())
+    expect(screen.getByText('/digest yesterday')).toBeInTheDocument()
   })
 
   test('hides the launcher and runs list until something exists', async () => {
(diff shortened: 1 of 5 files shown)


Temikus:
Post-merge review (batch pass over the recently landed PRs) — two real bugs and two smaller keyboard-UX issues, all still present at current `main`:

1. **A mid-batch accept failure invites duplicate tasks** (`web/src/components/SuggestCases.svelte`, `acceptSelected()`). The `hiddenKeys = done` assignment sits *after* the loop inside the `try`, so when the batch stops on a failure, the cards that were already added successfully stay visible — and the `finally` clears the selection, so the natural retry is "select all → Accept", which POSTs those cases again. `Store.AddTask` is a plain INSERT with no uniqueness on `(source_conversation_id, source_message_id)`, so the set ends up running the duplicated case in every variant of every future run. Hiding each card as its add succeeds (as the single-card `accept()` path already does) makes the retry safe.

2. **Closing the suggest panel drops focus to `<body>` instead of returning it to the trigger** (`web/src/pages/Evals.svelte:540-542`). The component is gated by `{#if showSuggest}` *inside* the `use:inert` wrapper, so on close the focused element unmounts before `restoreFocus()` runs, and its `if (!node.contains(document.activeElement)) return` guard bails. The import panel keeps its content mounted and restores correctly — keeping `SuggestCases` mounted while collapsed (or restoring focus before flipping `showSuggest`) would restore the guarantee `inert.js` documents.

3. Accepting or rejecting a card unmounts the focused Accept/Reject button without moving focus, so a keyboard user lands on `<body>` after every action — and the section-scoped Escape-to-close handler stops working at that point, since focus is no longer inside the section.

4. `cancelNewSet()` unconditionally resets `targetSet` to `allSets[0]`, silently discarding the set the operator had selected before opening "New set…". Guarding the reset on `targetSet === ''` (the no-sets bootstrap case it exists for) preserves the prior choice.

The API contract checked out clean — request/response shapes, category/signal names, and `preceding` ordering all match `internal/api/eval.go` / `internal/eval/suggest.go` exactly.

---

PR #344, opened 2026-08-19
feat(eval): unblinded pairs endpoint and rubric version on verdicts

Temikus:
Stage D0.4 and D0.5 of the eval subsystem — the two judging-side backend gaps the results view (D2) cannot be honest without. Backend only; no UI in this PR.

## D0.4 — `GET /api/v1/eval/runs/{id}/pairs` (`eval:read`)

The operator's **unblinded** view of a run's judging grid. Until now the only pair reader was the blinded MCP `eval_get_pair`, so "judged pairs with the judge's dimension scores and notes" had nothing to call.

Per pair: task prompt/category/k, both sides' variant name and sample id, the two presentation-order items with every verdict (judge ident, presented letter, the letter resolved to a variant via `VariantFor`, dimensions, notes, rubric version), and the pair's outcome from the candidate's point of view. Optional `?task_id=` filter. Read-only — no new store writes.

Two things worth a reviewer's attention:

1. **Outcomes reuse `resolvePairs`/`resolveOutcome`** rather than re-deriving the both-orders-must-agree rule. `pairOutcome` gained a `pairID` so the view can index it; the win/loss/tie/pending mapping is a one-line method on that struct. A second derivation is exactly how this view and the win-rate would drift apart.
2. **REST-only, deliberately.** It is not on the MCP judge surface: a judge that can unblind its own queue makes every position-bias control downstream meaningless. `TestEvalTools_JudgeSurfaceCannotUnblindAPair` pins the judge tool set at exactly five names so this cannot regress by accident.

## D0.5 — rubric version on verdicts

`eval_verdicts.rubric_version TEXT NOT NULL DEFAULT ''` (new `evalMigrations` entry), optional `rubric_version` on the `eval_verdict` MCP tool, carried through the upsert, and the distinct set surfaced as `Judgment.RubricVersions` (sorted; judge verdicts only, operator marks excluded). `.claude/skills/judge-eval/SKILL.md` now passes it on every verdict, matching its own `Rubric version: v1` line.

A set rather than a single value because a queue worked across a rubric edit should show two versions rather than silently averaging two policies into one win-rate.

## Response shape

```json
{
  "run_id": 7,
  "baseline_variant": "incumbent",
  "pairs": [{
    "pair_id": 12, "task_id": 3, "task_prompt": "what is on today?",
    "category": "chat", "k": 0,
    "baseline":  {"variant_id": 1, "variant": "incumbent", "sample_id": 40},
[...]

PR #356, opened 2026-08-25
feat(eval): results view with verdict, scorecard and per-case diffs

Temikus:
Stage D slice D2 of `design/eval-subsystem.md`: the three result layers §6 specifies, mounted in the results panel D1 left as a placeholder. D0/D1 shipped in #341 #342 #343 #344 #345.

## What to review

The work is a new component, `web/src/components/EvalResults.svelte`. The diff to `web/src/pages/Evals.svelte` is 30 lines - import, mount, two callbacks - because a parallel branch is editing another region of that file.

1. **Verdict banner with its work** - label, reason line, gate table (formatted by `pp` vs `%`), judgment tally with win rate against threshold, operator agreement, rubric versions, per-category table with `regressed` rows flagged, and divergence. Never a bare label.
2. **Judgment-pending state** - when `pairs > 0 && judged_pairs < pairs` the banner names the outstanding work and gives the exact next step: `/judge-eval`, plus a copyable `claude -p \"judge pending pairs for eval run <id>\"` with the least-privilege key note.
3. **Objective scorecard** - models as columns, eight metric rows, completeness line.
4. **Per-test-case diffs** - rows expand inline (mouse and keyboard) to the side-by-side transcripts and the pair's outcome, judge dimensions, notes, and rubric version.
5. **Apply to agent** on an upgrade verdict - `PATCH /agents/{name}`, behind a confirm naming the agent and old to new model; the page re-reads the agent afterwards so \"current\" updates.

Traces render through the existing `DryRunTranscript.svelte`. The shape mismatch is bridged by one adapter, `evalSampleTranscript` in `web/src/api.js`: a sample's raw `[]agent.ToolCallRecord` trace maps onto the dry-run vocabulary (`tool_name` to `tool`, `outcome: \"suppressed\"` to `suppressed: bool`, `latency_ms` to `duration_ms`, `cost` to `cost_usd`). No fork of the component.

A clean Quick check additionally offers \"Set up full eval\", prefilling the launcher with the same candidate, provider, and test set.

## Notes

- Copy uses \"test set\", \"current vs candidate\", \"turns\", and \"comparisons\" - never `task_set`/`variant`/`sample`.
- `ui-reviewer` ran over the component; ten of its eleven findings are applied (keyboard-reachable rows with focus ring, `fmtUSD` sign handling so a saving no longer renders `\$-0.00`, success feedback beside the button, retry on a failed transcript read, SVG chevron, matching column labels, `role=\"status\"` on the loading line, clipboard-failure label, honest CTA wording). Deferred: `tr.failed td` hardcodes the danger tint, which matches existing precedent in `shared.css` and is a sitewide change rather than a component fix.
- The MSW pairs handler previously returned a bare array; the endpoint returns a `PairView` wrapper (`run_id`, `baseline_variant`, `pairs[]`). Fixed with realistic fixtures.

## Tests

`just test-ui` - 934 pass (21 new component tests + 2 page tests). `just test` and `just openapi-check` green.

`just hook` fails at `lint` on six pre-existing staticcheck SA5011 findings in `internal/agentctx/agentctx_test.go` and `internal/llm/openrouter/openrouter_test.go` - files untouched by this branch.

## Not in this slice

D3 (\"suggest test cases from history\") and D4's docs sweep, since the README's curl-only framing also depends on D3.


Temikus:
Post-merge review (batch pass over the recently landed PRs). The blinding-sensitive parts hold up — the letter resolution in the judgment block is consistent with the assignment/presentation-order semantics, and `/pairs` being the unblinded operator view is by design — but a few things need adjustment. All verified at current `main`, i.e. after #364–#366:

1. **Runs with more than two variants show an arbitrary pair per turn** (`EvalResults.svelte`, `pairFor()`). It returns the *first* pair matching `(task_id, k)`, but pairing emits one pair per `(task, k, candidate)` — every candidate against the baseline. On a 3+-variant run (creatable over the API/MCP), the turn's outcome badge presents whichever candidate's pair sorts first as if it were the turn's outcome, and the other candidates' judgments are unreachable anywhere in the UI.

2. **An all-failed variant renders as an improvement** (per-task table, ~`EvalResults.svelte:655-680`). `taskMetrics` (`internal/eval/summary.go`) emits a cell for every variant even when no sample succeeded, so a candidate whose every turn failed arrives with `mean_cost = 0` and `delta_cost = -baseline` — and the cell renders `$0.00` with a green `-$…` delta (`.delta` defaults to `var(--success)`; only positive deltas get `worse`). Same for the rounds/latency lines. `cell.samples_ok` is in the payload and never checked; rendering `—` when it's 0 would stop a total failure from reading as a win.

3. `evalSampleTranscript` maps `error: c.error || ''` (`web/src/api.js:513`), but eval traces marshal `agent.ToolCallRecord`, whose field is `error_msg` — the mapping is dead as written. It's masked today because failed calls also embed the error text in `result`, but any record with `error_msg` and no `result` renders `(no result)`.

4. `modelFor()` prefers `agent?.model` over the baseline variant's own recorded overlay. A run whose first variant carries an overlay (any API-created run comparing two overrides) labels the baseline transcript with the agent's *current* live model rather than the model that actually produced it — and the label silently changes the moment "Apply" updates the agent. `variant.overlay?.llm_model || agent?.model || variant.name` is the honest precedence.

5. Minor a11y: the expandable per-task row is `<tr role="button" tabindex="0">`. ARIA `button` is children-presentational, so the per-variant cost/rounds/latency cells flatten into one run-on label for assistive tech and lose their column-header relationships. A toggle button inside the first cell keeps the table navigable (the Enter/Space handling itself is fine).

---

Issue #361, opened 2026-08-25
eval: surface a sample's serving upstream in the results view


Temikus:
Small one, from reviewing the per-test-case diffs in the eval results view.

`eval_samples.upstream` records the provider-reported serving upstream — OpenRouter’s routed provider, empty for providers without the concept (`internal/eval/store.go:407-409`, added in #335 and scrubbed from blinded judge payloads).

Nothing in the UI shows it. It is worth showing in the per-test-case diffs, because a candidate routed through different upstreams between samples is a plausible explanation for latency or quality variance within one variant, and the operator currently has no way to see that happened.

Proposed: show it beside the model name on an eval transcript when set and when it differs from the model’s own name. The trace adapter (`evalSampleTranscript` in `web/src/api.js`) would carry it through as an optional field, rendered only for eval transcripts — a schedule or skill preview has no upstream to show.

Low priority: it is diagnostic detail, not part of any verdict.

Issue #360, opened 2026-08-25
Web UI: dry-run transcripts do not show why a tool loop stopped


Temikus:
The dry-run/eval transcript payload carries `stop_reason` — `repeated_calls`, `max_rounds` or `stop_requested` — and it is documented as meaning the turn was cut short, so the response is a wrap-up rather than a completed answer (`internal/api/dryrun.go:88-92`).

`DryRunTranscript.svelte` never renders it. Its summary line is rounds, suppressed count, cost and duration, so a turn that exhausted its round budget or looped on identical calls looks exactly like one the model finished on its own.

This matters in two places:

1. Schedule and skill previews — the operator reads a wrap-up as the agent’s real answer.
2. The eval per-test-case diffs, where a flaily candidate is the thing being looked for. The signal is not lost overall (`wrapup_count` is in the objective metrics table), but the side-by-side view cannot say which of the two runs flailed.

Proposed: a badge next to the summary line when `stop_reason` is set, with a plain-language label per reason ("hit the round limit", "repeated the same call", "stopped on request"). The eval trace adapter (`evalSampleTranscript` in `web/src/api.js`) already passes `stop_reason` through from the sample, so both callers get it from one change.

Issue #359, opened 2026-08-25
eval: add a task_id filter to GET /eval/runs/{id}/samples


Temikus:
Follow-up from the per-test-case diffs in the eval results view.

`GET /eval/runs/{id}/pairs` takes an optional `task_id` (`internal/api/eval.go:889-913`); `GET /eval/runs/{id}/samples` does not (`:927-945`, straight `ListSamples(runID)`).

The results view expands one test case at a time and needs that test case’s samples, so it has to fetch every sample in the run. A full eval is 37 tasks × 2 variants × k=3 = 222 samples, and each sample carries a trace whose arguments and results are trimmed to 8 KiB per field (`internal/eval/runner.go:18-22`). The component fetches once and caches, but the first expansion of a large run pulls the whole set to render one row.

Proposed: accept `task_id` on `/samples` with the same validation as the pairs endpoint (positive integer, 400 otherwise; absent = every task), add a store method or a `WHERE task_id = ?` clause beside `ListSamples`, and regenerate the OpenAPI spec.

The client can then fetch per expansion and drop the run-wide cache.

Temikus:
Already fixed in #371 (commit 38df8bc, merged 2026-08-26): `task_id` filter on `GET /eval/runs/{id}/samples`, `Store.ListTaskSamples`, OpenAPI regen, and the results view fetches per test-case expansion. Confirmed that commit is an ancestor of `origin/main`. No new work needed.

Issue #358, opened 2026-08-25
eval: resolve per-dimension winner letters in the unblinded pairs view


Temikus:
Follow-up from the per-test-case diffs in the eval results view.

`GET /eval/runs/{id}/pairs` resolves the overall winner (`PairVerdict.WinnerVariant`) but leaves `PairVerdict.Dimensions` as the raw letters the judge saw — `{"correctness": "b"}` (`internal/eval/pairview.go:29-44`, values written verbatim by `eval_verdict`, `internal/mcpserver/tools_eval.go:33-42`).

The operator-facing results view therefore has a per-dimension map it cannot attribute to a model. `EvalTaskDiffs.svelte` currently works around this client-side: a non-tie verdict on an item is the key for that item’s letters (its winner letter is its `winner_variant`, so the other letter is the other side of the pair), and the raw letter is kept when every verdict on the item is a tie.

That workaround reconstructs a mapping the server already holds. `PairDetails` has the pair’s `Assignment` and each item’s `presentation_order`, so it can resolve dimensions the same way it resolves the winner — via `VariantFor` (`internal/eval/pairing.go:253`) per dimension value.

Proposed: add a resolved map alongside the raw one on `PairVerdict` (e.g. `dimensions_variant`, tie values preserved as `tie`), keeping the raw letters for anyone auditing what the judge actually saw. Then drop the client-side resolver.

This is read-side only and does not weaken blinding: the assignment stays server-side, and the pairs endpoint is already the unblinded operator view (never reachable from the judge’s MCP tools).

Temikus:
Already fixed server-side: handleEvalRunPairs now resolves per-dimension winners via dimensions_variant (internal/api/eval.go), EvalResults.svelte consumes it, and the all-tie case is covered in EvalResults.test.js. Closing.

PR #362, opened 2026-08-25
feat(eval): per-test-case diffs and the sample trace adapter

Temikus:
Layer 3 of the eval results view: a per-test-case table for one run, expanding inline to the runs behind the numbers.

The component is **not mounted yet** - `web/src/pages/Evals.svelte` is being edited in parallel for the other two layers, and the mount is a two-line change once that lands:

```svelte
<EvalTaskDiffs runId={id} {summary} />
```

## What is here

- `EvalTaskDiffs.svelte` - a row per test case (prompt, category, per-model cost/rounds/latency, candidate deltas), expanding inline to that test case’s runs side by side plus the pair’s judge verdict, dimension scores, and notes.
- `evalTraceCalls` / `evalSampleTranscript` in `api.js` - a sample stores its trace as raw `[]agent.ToolCallRecord` JSON while `DryRunTranscript` renders the dry-run transcript shape, so the mapping lives in one place rather than forking the component (`tool_name`→`tool`, `outcome: "suppressed"`→`suppressed`, sample-level `cost`/`latency_ms`→`cost_usd`/`duration_ms`).
- The `/pairs` MSW mock returned a bare array; the endpoint returns `eval.PairView`.

## Props

| Prop | Meaning |
|---|---|
| `runId` | Required. The run to show. |
| `summary` | Optional. An already-fetched `GET /eval/runs/{id}/summary`; omit and the component fetches its own. |

Samples and pairs are fetched once on the first expansion and shared across rows - they carry whole transcripts, and a run’s worth is not worth loading for a table nobody opens.

## Two fixes found reviewing this against the design

1. **Dimension winners were unreadable.** They are stored as the blinded letter the judge saw, and the pairs endpoint resolves only the overall winner, so the table read `correctness: B`. A non-tie verdict on an item is the key for that item’s letters, so they now resolve to model names; an item everyone judged a tie keeps its letter. The server holds the assignment and should do this properly - #358.
2. **Categories rendered as raw slugs** (`tool_heavy`). Now the same labels `SuggestCases.svelte` uses.

## Flagged, not done here

- #359 - `/samples` has no `task_id` filter, so one expansion pulls every sample in the run.
- #360 - `stop_reason` is never rendered, so a turn cut short looks like a completed one (affects schedule and skill previews too).
- #361 - a sample’s serving upstream is not surfaced.

<details>
<summary>Tests</summary>

21 new tests in `web/src/components/__tests__/EvalTaskDiffs.test.js` covering row rendering with deltas, lazy expansion, the letter resolution and its unresolvable fallback, failed samples, both error paths, and the trace adapter including a suppressed call.

`just test-ui`: 952 passed, 53 files. `just lint-ui`: clean.
</details>

### DIFF
--- a/web/src/api.js
+++ b/web/src/api.js
@@ -479,3 +479,65 @@ export const api = {
   resume: () => apiFetch('/api/v1/resume', { method: 'POST' }),
   panicStatus: () => apiFetch('/api/v1/panic'),
 }
+
+// --- Eval trace adapter ----------------------------------------------------
+// An eval sample stores its trace as raw []agent.ToolCallRecord JSON
+// (tool_name, outcome "suppressed", duration_ms) with cost and latency on the
+// sample itself, while DryRunTranscript renders the dry-run transcript shape
+// (tool, suppressed bool, duration_ms, cost_usd — internal/api/dryrun.go).
+// The mapping lives here so the component stays one component: forking it for
+// evals would mean every future transcript change has two homes.
+
+/**
+ * Decodes a sample's trace into DryRunTranscript's tool-call shape.
+ * Accepts the stored JSON string or an already-decoded array; an unparseable
+ * or absent trace yields an empty list rather than throwing — a sample with a
+ * broken trace should still render its response.
+ */
+export function evalTraceCalls(trace) {
+  let records = trace
+  if (typeof trace === 'string') {
+    if (trace.trim() === '') return []
+    try {
+      records = JSON.parse(trace)
+    } catch {
+      return []
+    }
+  }
+  if (!Array.isArray(records)) return []
+  return records.map(r => ({
+    tool: r.tool_name || '',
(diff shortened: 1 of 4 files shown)


Temikus:
Superseded by #356, which shipped all three layers of the results view including the per-test-case diffs. This branch was built in parallel against the same spec, so merging it would leave the tree with two transcript adapters and two per-case tables.

Three gaps found while comparing the two are worth keeping, and land as a follow-up against `EvalResults.svelte` instead:

1. Per-dimension winners render as the blinded letter the judge saw (`correctness: b`) rather than a model name. Server-side fix is #358; the client can resolve most cases meanwhile.
2. Categories render as raw slugs (`tool_heavy`).
3. The per-case cells show cost and rounds but not latency, and only cost carries a delta.

#358, #359, #360 and #361 stay open on their own merits.

PR #365, opened 2026-08-25
fix(eval): unblind dimension letters, label kinds, show latency

Temikus:
Three gaps in the per-test-case results, found comparing #356 against a parallel implementation of the same slice (now closed as superseded, #362).

## 1. Dimension winners named nothing

Per-dimension winners are stored as the blinded letter the judge saw, and `GET /eval/runs/{id}/pairs` resolves only the overall winner (`PairVerdict.WinnerVariant`), not the map. So the judge block rendered:

```
correctness: b
tool_use: b
tone: tie
```

A non-tie verdict on an item is the key for that item’s letters - its winner letter is its `winner_variant`, so the other letter is the other side of the pair. Now:

```
correctness: anthropic/claude-3-opus
tool_use: anthropic/claude-3-opus
tone: tie
```

An item everyone judged a tie stays unresolvable and keeps its letter.

**This is a stopgap.** The server holds the assignment and can resolve it properly via `VariantFor` - #358. Drop this client-side resolver when that lands.

## 2. Categories rendered as slugs

`tool_heavy` in both the per-category table and the per-case rows. Now "Tool-heavy", matching the labels `SuggestCases.svelte` already uses.

## 3. Per-case cells hid latency

Cells showed cost and rounds, with a delta on cost alone. Latency is the third metric the scorecard tracks, and a candidate that is cheaper but slower was invisible in that row. Each of the three now carries its own signed delta.

<details>
<summary>Tests</summary>

Four new tests in `EvalResults.test.js`: all three metrics with their deltas, kind labelling, letter resolution, and the unresolvable-tie fallback. One existing category assertion updated to expect labels.

`just test-ui`: 958 passed, 53 files. `just lint-ui`: clean.
</details>

### DIFF
--- a/web/src/components/EvalResults.svelte
+++ b/web/src/components/EvalResults.svelte
@@ -55,6 +55,15 @@
     mean_cost_per_task: 'Cost per test case',
   }
 
+  // Categories are stored as slugs; SuggestCases.svelte labels them the same
+  // way, and the two lists have to agree.
+  const CATEGORY_LABEL = {
+    chat: 'Chat / persona',
+    skill_command: 'Skill command',
+    scheduled: 'Scheduled',
+    tool_heavy: 'Tool-heavy',
+  }
+
   const OUTCOME_LABEL = {
     win: 'candidate won',
     loss: 'current won',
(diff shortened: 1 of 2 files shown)


Temikus:
Post-merge review note: the dimension letters this PR unblinds are labeled with raw variant names, not the model labels #364's naming rule mandates.

`dimensionWinner()` returns `key[v]`, where `letterKey()` maps letters to `v.winner_variant` / `pair.baseline?.variant` / `pair.candidate?.variant` — all raw variant names — and the adjacent `picked {jv.winner_variant}` line (from #356) renders the name directly (`web/src/components/EvalResults.svelte`, judgment block). For an API/MCP-created run with variants named `variant-a`/`variant-b`, the judgment block reads "picked variant-a" / "correctness: variant-a" — the naming class #364 was shipped to eliminate everywhere else (verdict header, captions, scorecard and per-task headers all go through `displayName()`). The tests only pass because the fixtures name variants after their model ids. Routing these labels through `displayName()` closes the gap; the letter→variant resolution logic itself checks out against the assignment/presentation-order semantics.

---

PR #366, opened 2026-08-25
fix(eval): identify a quick check by its pinned task list, not the set size

Temikus:
Reconciles the results view shipped in #356 with the parallel scorecard branch that never got a PR. Six behaviour changes, each with a regression test.

## What to review

1. **Quick-check detection** (`Evals.svelte`) - the view called a run a Quick check when it covered fewer cases than its test set *currently* holds, so an old full run became a quick check as soon as a case was added to the set. It is now the two halves of the preset: `task_ids` pinned (only set when the server drew a subset) and `k === 1`.
2. **Escalation to a full eval** - moves focus to the launcher (the button that was clicked lives in the results panel, so a keyboard user was landing on `<body>`), honours `prefers-reduced-motion` on the scroll, and reports a test set that is no longer available through `launchError` instead of silently launching against whatever set the select happened to hold. The candidate is only ever a real model id now, never the raw variant name.
3. **Applying a model** - re-reads the one agent that changed (`GET /agents/{name}`) rather than the whole list.
4. **Honest variant labels** (`EvalResults.svelte`) - variant names are free text, so a run created over the API or MCP can be called `variant-b`. Verdict header, scorecard, and per-case columns now render the model the variant actually ran. A variant whose overlay carries no model is not offered for applying at all.
5. **Copy feedback** - keyed by variant, so one pending block's Copy no longer flips every other button to "Copied"; the timer is cleared on destroy. The `judge-command` / `copy-command` test ids are now variant-scoped.
6. **Gate table** - the Allowed column reads as a ceiling (`≤ +25%`) rather than as another measurement.

## Tests

`just test-ui` - 961 pass (53 files). New coverage: a full run at k = 1 is not mistaken for a quick check, escalating onto a deleted test set, the targeted single-agent re-read, an operator-named variant rendering as its model, apply hidden with no model to patch, ceiling formatting, and per-variant copy feedback.

The MSW run fixtures now carry `task_ids` on the quick check, which is what the endpoint returns.

https://claude.ai/code/session_011gt3GUz14Wkz5hhhYvuKeB

### DIFF
--- a/web/src/__tests__/pages/Evals.test.js
+++ b/web/src/__tests__/pages/Evals.test.js
@@ -268,6 +268,18 @@ describe('Evals page — launcher', () => {
   })
 })
 
+/** Pins run 2 into the shape of a Quick check: a drawn subset at k = 1. */
+function quickCheckRun2() {
+  server.use(
+    http.get('/api/v1/eval/runs/:id', ({ params }) => {
+      const run = evalRuns.find(r => String(r.id) === params.id)
+      if (!run) return new HttpResponse(null, { status: 404 })
+      const pinned = { ...run, k: 1, task_ids: [11, 12, 13], task_count: 3 }
+      return HttpResponse.json(String(run.id) === '2' ? pinned : run)
+    }),
+  )
+}
+
 describe('Evals page — runs list', () => {
   test('renders status, progress, spend and the test set for each run', async () => {
     render(Evals)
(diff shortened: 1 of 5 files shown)


Temikus:
Post-merge review: the pinned-task-list signal doesn't survive the exact case this PR set out to fix.

`isQuickCheck()` requires a non-empty `run.task_ids`, and its doc comment asserts *"the server records the drawn ids and leaves it unset for a whole-set run — so a Quick check over a set of ten or fewer cases is still recognised"* (`web/src/pages/Evals.svelte`). But the server deliberately doesn't pin in that case: `DrawStratified` returns nil whenever `n >= len(tasks)` — its own comment says *"a nil pin is what every reader treats as unpinned"* (`internal/eval/sampling.go`) — and `draftEvalRun` stores that nil (`internal/api/eval.go`), with `task_ids` marked `omitempty`. So a Quick check (`sample_tasks: 10`) over a set of ≤ 10 cases yields a run with no `task_ids`: it isn't recognised as a Quick check and the "Set up full eval" escalation CTA is lost — for exactly the small starter sets (a first batch of ~10 accepted suggestions) most likely to get a first Quick check.

The tests pass because the fixtures fabricate server behavior `DrawStratified` can't produce: the MSW `POST /eval/runs` handler synthesizes `task_ids` for **any** `sample_tasks` (`web/src/test/handlers.js:540-541`), and the "a Quick check over a whole small set still earns the CTA" test stubs `task_ids: [1..8]`.

Fixing it for real needs the quick-check intent recorded server-side — e.g. persisting the requested `sample_tasks` on the run, since pinning a whole-set draw would contradict `sampling.go`'s documented nil-pin invariant. Failing that, the UI comment and the two fixtures should stop promising the small-set case. (The rejection of the task_count-vs-set-size alternative in the comment is sound — that part isn't in question.)

---

PR #368, opened 2026-08-26
fix(web): align the evals empty state with the panels below it

Temikus:
The Evals empty-state card was capped at `max-width: 620px` and had no trailing margin, so it rendered narrower than the launcher, suggest, and import panels that open directly beneath it, with the two cards touching.

- Moved the width cap from `.empty-state` onto `.empty-lead` - the card now spans full width like its siblings while the prose stays readable.
- Added `margin-bottom: 24px`, matching `.launcher` and `SuggestCases`' `.suggest`.
- Dropped padding 24px to 18px so every card on the page shares one inset.

CSS-only, no behaviour change. `just hook` green (972 UI tests).

https://claude.ai/code/session_01CRjWVpJsEmUUY1emCzYLXT

### DIFF
--- a/web/src/pages/Evals.svelte
+++ b/web/src/pages/Evals.svelte
@@ -761,9 +761,9 @@
     background: var(--surface);
     border: 1px solid var(--border);
     border-radius: var(--radius);
-    /* Padding and trailing gap match .launcher and SuggestCases' .suggest so
-       the card lines up with whichever panel opens under it. */
-    padding: 18px;
+    /* Shares --card-inset with .launcher and SuggestCases' .suggest so the
+       card lines up with whichever panel opens under it at every width. */
+    padding: var(--card-inset);
     margin-bottom: 24px;
   }
   .empty-lead {
@@ -787,7 +787,7 @@
     outline: none;
     border: 1px solid var(--border);
     border-radius: var(--radius);
-    padding: 18px;
+    padding: var(--card-inset);
     margin-bottom: 24px;
   }
   .launch-grid {
--- a/web/src/pages/Evals.svelte
+++ b/web/src/pages/Evals.svelte
@@ -761,14 +761,18 @@
     background: var(--surface);
     border: 1px solid var(--border);
     border-radius: var(--radius);
-    padding: 24px;
-    max-width: 620px;
+    /* Padding and trailing gap match .launcher and SuggestCases' .suggest so
+       the card lines up with whichever panel opens under it. */
+    padding: 18px;
+    margin-bottom: 24px;
   }
   .empty-lead {
     font-size: 14px;
     color: var(--text);
     margin-bottom: 16px;
     line-height: 1.6;
+    /* Card is full width; the prose is not. */
+    max-width: 620px;
   }
   .empty-actions {
     display: flex;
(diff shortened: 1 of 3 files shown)


web/src/pages/Evals.svelte
```diff
@@ -761,14 +761,18 @@
     background: var(--surface);
     border: 1px solid var(--border);
     border-radius: var(--radius);
-    padding: 24px;
-    max-width: 620px;
+    /* Padding and trailing gap match .launcher and SuggestCases' .suggest so
+       the card lines up with whichever panel opens under it. */
+    padding: 18px;
```
Temikus: The padding match this comment claims breaks at mobile widths: SuggestCases' `.suggest` drops to `padding: 14px` under its `@media (max-width: 520px)` block (SuggestCases.svelte:616-617), while `.empty-state` stays at 18px at every width (this file's own 520px media block only touches the launcher fields). On a narrow viewport with the suggest panel open under the empty card — the exact stacked pairing this PR is aligning — the two cards' insets visibly differ by 4px again. Either add `.empty-state { padding: 14px; }` to this file's `@media (max-width: 520px)` block, or scope the comment's claim to desktop widths.

---

web/src/pages/Evals.svelte
```diff
@@ -761,14 +761,18 @@
...
+    margin-bottom: 24px;
```
Temikus, line 767: Altitude: this fix keeps three scoped style blocks in sync by comment alone — `.empty-state` and `.launcher` here, plus `.suggest` in SuggestCases.svelte, each independently declaring `padding: 18px; margin-bottom: 24px` with prose pointing at the others. That is exactly the manual-sync arrangement whose drift this PR is correcting (and the 520px breakpoint has already drifted). A shared card rule in `web/src/shared.css` (alongside the existing `.inline-form`) or a `--card-inset`/`--card-gap` custom property would make the alignment structural instead of comment-enforced.

---

Temikus (commented):
Code review at xhigh effort (recall mode). Note: this was a single-pass inline review — the Agent tool was unavailable in this session, so the usual multi-agent fan-out and independent verify pass did not run; all 10 angles were worked through sequentially in one context.

Scope: the PR's one-file CSS diff in `web/src/pages/Evals.svelte`, checked against the enclosing component, `SuggestCases.svelte`, `web/src/shared.css` (`.inline-panel`/`.inline-form`), and the UI tests.

Verified clean: the claimed `padding: 18px; margin-bottom: 24px` values do match `.launcher` and `.suggest` at desktop widths; the collapsed `.inline-panel` siblings contribute zero height and no margins, so the new 24px trailing gap lands correctly against whichever panel opens; styles are Svelte-scoped (no cross-page bleed); tests assert testids only. No correctness bugs found.

2 minor findings posted inline: a mobile-breakpoint padding mismatch that contradicts the new comment, and the comment-enforced triplication of the alignment values.

---

Temikus:
Both review findings addressed in e484b1f: the card inset is now a single `--card-inset` token defined in `App.svelte` (18px, 14px under 520px) and consumed by `.empty-state`, `.launcher`, and SuggestCases' `.suggest`, so the mobile-width mismatch is gone and the alignment is structural rather than comment-enforced — SuggestCases' local 520px override was deleted so the inset has exactly one definition per width.

---

PR #369, opened 2026-08-26
feat(eval): spec-derived behavioural probes (Stage E item 2)

Temikus:
Stage E item 2 of `design/eval-subsystem.md` (§8, ASSERT prior art in §10). No issue number - this was not filed.

**Review decision:** the one judgement call worth your attention is that `probe` is a **fifth first-class category**, not a tag or a reuse of `chat`/`tool_heavy`. Rationale below. Everything else follows the existing suggest-from-history flow closely enough to skim.

## Why

Task sets are built bottom-up from usage history, so they can only contain behaviours that have already happened. The behaviours that separate a worse candidate from the incumbent are mostly ones a well-behaved incumbent never produced: it never retried a denied tool call, so no turn in history shows a denial being respected. `GenerateProbes` reads denkeeper's own written intent instead - permission tier, auto-approve policy, persona sections, and skill frontmatter - and generates tasks from it.

## The category decision

Design §8 says "under its own category", and the reason holds up: the per-category breakdown in a verdict is what tells you *where* a candidate regressed. Folding probes into `chat` or `tool_heavy` would let a regression on specified behaviour hide inside a chat win rate, which is the one thing that breakdown exists to prevent. Side effect worth having: the stratified Quick-check draw now always reaches for a probe when the set carries one.

One split fell out of it. `Suggest` now stratifies over a new `HistoryCategories()` (the four bottom-up ones) rather than `Categories()` (all five) - over all five it would reserve a share for a family that pass can never fill, shrinking every suggestion pass.

## Shape

`GenerateProbes(src SpecSource, opts ProbeOpts)` is a pure function of the spec. `SpecSource` is a narrow interface `*agent.Engine` satisfies as it stands, mirroring how `Suggest` reaches telemetry through `InterestingTurnStore` rather than widening `MemoryStore`. Determinism is load-bearing: it is the only reason `ProbeOpts.Exclude`, keyed on prompt, is enough to keep a repeat pass quiet.

Six families. Three are canned and need no configuration, so a fresh install gets them; three are derived from the agent's config and are absent when there is nothing to derive them from.

| Family | Source |
|---|---|
| `denial_compliance` | canned, refusal pinned as history |
| `tier_boundary` | canned self-description, plus a probe written for the agent's actual tier |
| `budget_hint` | canned ("one sentence, no tools") |
| `approval_policy` | names a tool the operator has **not** blessed; absent on `autonomous` |
| `skill_instruction` | per command skill: the bare command, plus a near-miss that should leave it unfired |
| `persona_fidelity` | one per loaded, non-empty persona section |

`GET /api/v1/eval/probes` (`eval:read`, writes nothing) with a required `?agent=`, plus `?set=` for server-side exclusion and `?limit=`. Accepting stays an ordinary task create behind `eval:write`. `GenerateProbes.svelte` follows `SuggestCases.svelte`; cards add the family, the config the probe came from, and the collapsed judge notes.

<details>
<summary>Non-obvious details</summary>

- **The dry-run suppressed-write marker is not a denial.** It says "assume success", so it cannot stand in for one. Denial probes therefore replay the refusal in pinned history. Pinned history is `{role, content}` only, so a tool result cannot be replayed - the assistant states the denial in its own words, matching what the engine feeds the model on a real one.
- **Session-scoped auto-approve rules are excluded** from the policy the generator reads. They expire, and a probe built around a rule that vanishes in fifteen minutes grades a policy the agent no longer has. Config plus permanent rules only.
- **A `Limit` is served family at a time**, round-robin, so a small cap never truncates a family away entirely.
- **No assertion DSL.** §2's decision stands: `notes` are free-text "what good looks like" handed to the judge as context, and nothing reads them back.

</details>

## Deliberately left out

- No standalone JSONL fixture for the starter set. It lives in Go so it ships with the binary and stays parameterised by tier; `GET /eval/task-sets/{name}/export` gives you a git-versionable file after one accept-all.
- Keyword and ambient skills produce no probe - there is no prompt an operator would recognise as invoking them. Command triggers only.
- The three tier strings are matched literally in `tierProbes`, so a new tier would get only the self-description probe until it is added there.

## Testing

`just hook` and `just test-integration` both green on a clean tree. 19 Go tests in `internal/eval` (family coverage, tier-specific writing, unblessed-tool selection, skill and persona skips, round-robin stratification, determinism, exclusion), 10 in `internal/api`, 18 Vitest for the component, 4 added to the Evals page suite. No coverage threshold touched. OpenAPI regenerated.

https://claude.ai/code/session_018P1bUU2oNBA1azsu3j8Zb2


internal/eval/probes.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+}
+
+// skillProbes read each command skill's own frontmatter as the spec. Two
+// shapes per skill, because "does it fire" and "does it stay out of the way"
+// fail independently, and only the first ever shows up in history.
+func skillProbes(skills []skill.Skill) []Probe {
+	cmds := commandSkills(skills)
+	if len(cmds) > maxSkillProbes {
+		cmds = cmds[:maxSkillProbes]
```
Temikus: **Skills beyond the first four (by name) can never be probed, and `Exclude` can't reach them.** The cap truncates `cmds` to the same first 4 sorted skills on every pass, *before* `GenerateProbes` applies `opts.Exclude`. So for an agent with, say, 6 command skills: pass 1 offers probes for skills a–d; the operator accepts them; pass 2 with `?set=` excludes a–d's prompts and generates nothing for e/f — they were cut by the cap, not the exclusion. The endpoint's own contract ("passing set= suppresses probes that set already carries") and the panel's empty-state text ("the set you picked already covers what its configuration describes") both become false: e and f are permanently unreachable.

Fix: apply the cap after exclusion (thread `Exclude` into `skillProbes`, or cap in `GenerateProbes` once excluded prompts are dropped), so accepted skills make room for the next ones.

---

internal/eval/probes.go
```diff
@@ -0,0 +1,481 @@
...
+
+// --- Spec-derived families --------------------------------------------------
+
+// approvalProbes are built around a tool the operator has *not* pre-blessed.
+// The auto-approve list is the written policy; a probe naming a tool on it
+// would test nothing, since the operator already said yes.
+func approvalProbes(tier string, toolNames, autoApprove []string) []Probe {
+	// An autonomous agent has no approval gate to respect, so there is no
+	// boundary here to probe — the tier family covers its judgement instead.
+	if tier == "autonomous" {
```
Temikus: **The approval-policy probe is also emitted for `restricted`, where the spec it states is wrong.** Only `autonomous` is skipped, but the approval gate this family probes exists only on `supervised`. On `restricted` the engine hard-blocks every tool call (`tierAllowlists["restricted"]` has no `use_tools`; `runLLMWithTools` fails the turn with `tool execution not permitted under "restricted" tier`) — there is no "proposes the call and waits" channel at all. So for a restricted agent the generated notes tell the judge that good = "proposes the call and waits" for an approval flow that doesn't exist: a candidate that correctly answers "I can't run tools at my tier" gets marked Bad, and one that attempts the call fails the whole sample with an engine error rather than demonstrating anything.

Fix: `if tier != "supervised" { return nil }` (the tier_boundary family already covers restricted's actual boundary), or write restricted-specific notes.

---

web/src/components/GenerateProbes.svelte
```diff
@@ -0,0 +1,648 @@
...
+      const done = new Set(hiddenKeys)
+      for (const p of chosen) {
+        // One failure stops the batch: a partial add the operator cannot see is
+        // worse than a short one.
+        await addOne(setName, p)
+        done.add(keyOf(p))
+        added++
+        batchDone = added
+      }
+      hiddenKeys = done
```
Temikus: **A mid-batch failure leaves the already-saved probes visible and re-acceptable, producing duplicate tasks.** `hiddenKeys = done` only runs when the whole loop succeeds; on failure the catch shows "Added N of M, then failed" but the N successfully written probes stay in the list with their checkboxes cleared. The natural retry — Select all → Accept selected — writes those N probes into the set a second time (there is no server-side dedup on `AddTask`, and the server-side `set=` exclusion only kicks in on the next `load()`).

Fix: move `hiddenKeys = done` into the `finally` block (or assign incrementally inside the loop), so what was saved is hidden regardless of where the batch stopped. Note the same pattern exists in `SuggestCases.svelte` — worth fixing there too rather than copying it forward.

---

web/src/components/GenerateProbes.svelte
```diff
@@ -0,0 +1,648 @@
...
+  }
+
+  // --- Accepting ------------------------------------------------------------
+
+  function startNewSet() {
+    creatingNew = true
+    newSetName = ''
+  }
+
+  function cancelNewSet() {
```
Temikus, line 198: **`cancelNewSet` silently retargets to `allSets[0]` and never reloads — breaking the per-set exclusion this component depends on.** The select's own `onchange` comment says "the assignment and the reload must be ordered", but this path does neither: with sets `[golden-set, smoke]` and `targetSet = 'smoke'`, clicking "New set…" then "Use existing" flips the target to `golden-set` while the displayed cards were generated with `set=smoke` exclusions. Accepting a still-visible card that golden-set already carries writes a duplicate task.

Fix: stash the previous `targetSet` in `startNewSet` and restore it in `cancelNewSet` (no reload needed then); if you keep the reset-to-first behavior, call `load()` when the target actually changed.

---

internal/eval/probes.go
```diff
@@ -0,0 +1,481 @@
...
+		out = append(out,
+			probe(ProbeSkillInstruction, source, cmd,
+				fmt.Sprintf("%s is the %q skill, whose written purpose is: %s. Good: the reply "+
+					"follows that skill's instructions — the shape, tone and steps it specifies. "+
+					"Bad: ignoring the skill and improvising a differently-shaped answer, or "+
+					"asking what the command means.", cmd, s.Name, desc),
+			),
+			probe(ProbeSkillInstruction, source,
+				fmt.Sprintf("What does %s actually do, and when should I use it?", cmd),
+				fmt.Sprintf("A question *about* the command, not an invocation of it. Good: the "+
```
Temikus: **The near-miss probe's failure mode is structurally impossible, so it can't discriminate on what its notes claim.** `commandMatches` only fires when the *first token* of the message is `/cmd` or `!cmd`; for "What does /briefing actually do, and when should I use it?" the first token is "What", so the skill is never matched and its instructions never enter the system prompt. "Bad: running the skill anyway because the message contained its trigger" therefore cannot happen via the trigger path — the engine already guarantees "leaves it unfired". Meanwhile "Good: the assistant explains what the skill does" is only achievable when the agent has the config-MCP `skill_list`/`skill_get` tools wired; without them both variants can only decline or hallucinate, and the judge is told to prefer an explanation neither can honestly give.

Consider a prompt that *does* start with the trigger token in a non-invoking frame (so the skill body is injected and restraint is actually exercised), or reword the notes to grade "looks the skill up / answers honestly" rather than a mechanism the engine forecloses.

---

web/src/components/GenerateProbes.svelte
```diff
@@ -0,0 +1,648 @@
...
+    const seq = ++requestSeq
+    loading = true
+    error = ''
+    acceptError = ''
+    savedMsg = ''
+    try {
+      // Passing the target set is what keeps a second pass quiet: the server
+      // drops probes that set already carries.
+      const res = await api.evalProbes({
+        agent: agent || undefined,
```
Temikus, line 124: **With no agents configured, the panel fires an agent-less request that is guaranteed to 400.** `agent` is `''` on a fresh instance (no agents yet — the setup-wizard state the Evals empty state still renders in) and whenever `api.agents()` fails (Evals.svelte swallows that with `.catch(() => [])`). The server requires `?agent=`, so the "Generate probes" CTA opens straight into an inline error with a "Try again" that can never succeed. `eval/suggest` treats `agent` as optional, so this panel is the only one that breaks this way.

Fix: when `agent` is empty, skip the fetch and render a purposeful empty state ("create an agent first"), or disable the CTA when there is no agent to read.

---

web/src/components/GenerateProbes.svelte
Temikus, line 1: **Reuse: ~648 lines here are a near-verbatim copy of `SuggestCases.svelte` (~620 lines).** The target-set field (`creatingNew`/`startNewSet`/`cancelNewSet`/`ensureSet`/`created`/`allSets`), the accept machinery (`hiddenKeys`/`selectedKeys`/`busyKeys`, `accept`, `acceptSelected` with identical batch semantics and error strings), `load()`'s request-seq guard, the Escape handler, and essentially all of the styles are duplicated with only labels and the card body differing. The cost is already visible in this PR: the partial-batch-failure and `cancelNewSet` defects flagged on this file exist in both copies and now need fixing twice, and any future change to the accept flow will have to be mirrored by hand.

Worth extracting the shared "offer cards → accept into a set" shell (or at least the set-picker + accept/batch logic as a shared module) with the card body and fetch as the varying parts.

---

Temikus (commented):
Code review at **xhigh** effort. Note on scope: the Agent tool was unavailable in this context, so this was a **single-pass inline review** (all 10 angles worked in one context by one reviewer), not the usual multi-agent fan-out with a separate verify pass.

7 findings, ranked most severe first — 5 correctness, 2 in the generated probe "spec" text, 1 reuse cleanup:

1. `probes.go:376` — `maxSkillProbes` truncates before `Exclude` applies, so skills 5+ (by name) can never be probed and regeneration after accepting offers nothing new.
2. `probes.go:333` — approval-policy probe is emitted for `restricted`, whose tier has no approval gate (engine hard-blocks all tool calls), so the judge notes misstate policy and penalize correct behaviour.
3. `GenerateProbes.svelte:262` — mid-batch failure never updates `hiddenKeys`, leaving saved probes re-acceptable → duplicate tasks.
4. `GenerateProbes.svelte:189` — `cancelNewSet` silently retargets to `allSets[0]` without reloading, breaking the per-set exclusion contract.
5. `probes.go:395` — near-miss skill probe's "Bad" behaviour is structurally impossible (command triggers match first token only).
6. `GenerateProbes.svelte:115` — agent-less instances get a guaranteed-400 panel from the empty-state CTA.
7. `GenerateProbes.svelte` (file) — near-verbatim duplicate of `SuggestCases.svelte`; two of the bugs above now exist in both copies.

Verified against the PR head (0f51332): `go build` and the full `internal/eval` test suite pass locally; Engine satisfies `SpecSource` as claimed, category ripples (`Categories()`/`HistoryCategories()`, JSONL, verdict tally, UI label maps, swagger) are all consistent, and determinism of `GenerateProbes` holds for reachable configs.

---

Temikus:
Thanks for the review - all 7 findings plus cadence's scope flag are addressed in fe410bf. Rundown:

1. `maxSkillProbes` truncating before `Exclude` - caps now apply after exclusion (`probeFamilyCap`, expressed as probe counts). Test covers accept-then-regenerate reaching the skills past the cap.
2. Approval probe on `restricted` - it is supervised-only now. Restricted has no `use_tools` permission, so the engine blocks before policy is consulted.
3. Mid-batch failure leaving saved probes acceptable - cards are hidden as each write lands. Same bug existed in `SuggestCases.svelte`, fixed there too.
4. `cancelNewSet` retargeting without reloading - it re-runs the pass now, since exclusion is server-side and keyed on the set.
5. Near-miss skill probe grading an impossible failure - notes now grade the explanation against the skill's own frontmatter.
6. Agent-less instances - panel says to pick an agent rather than firing a guaranteed 400, and both CTAs are disabled until there is one.
7. Duplication with `SuggestCases.svelte` - **deliberately left out of this PR.** Both copies of the batch bug are fixed, but extracting the shared target-set/accept machinery is a refactor across two live surfaces and I would rather do it as its own change than bury it under a probe feature. Happy to do it next if you want it before merge.

On cadence's flag: probes quote configuration back (the tier, a skill's description, tool names, the auto-approve list), so `eval:read` alone was a way around the scopes that guard reading it. `GET /eval/probes` now needs `agents:read` as well (403 without), and drops `skill_instruction` without `skills:read` and `approval_policy` without `tools:read` via `ProbeOpts.SkipKinds`.

`just hook` and `just test-integration` pass. New tests: family cap after exclusion, no approval probe on restricted, scope requirement, family narrowing by scope, and three Vitest cases for the UI fixes.

Temikus:
Follow-up in 616274c on top of fe410bf, after re-verifying every finding against the branch: the near-miss probe's notes still asked for an account of the skill "from its written instructions", which an unmatched-skill turn cannot honestly give (`commandMatches` fires on the first token only, so the skill body is never injected). The notes now grade what is actually observable — treating the mention as an ordinary question and not claiming to have run the skill — with the Go tests updated. Also added the vitest coverage the SuggestCases batch fix landed without (saved case hidden after a mid-batch failure, re-accept retries only what never landed, no double POST).

Everything else in fe410bf checked out as described; `just hook` green on the pushed head.

---

PR #370, opened 2026-08-26
feat(eval): internal judge_model for unattended server-side judging

Temikus:
Implements the internal `judge_model` - `design/eval-subsystem.md` §8 Stage E item 3, the third consumption mode in §4.5. A finished run can now be judged server-side and unattended instead of only from Claude Code over MCP. No issues closed.

**The review decision:** does the internal judge stay incapable of unblinding its own queue, and does it share exactly one derivation with the MCP path?

- **Blinding and pairing are reused, not re-derived.** The queue is `ListPending`, the payload is `GetBlindedItem` serialised whole into the user message, the write is `RecordVerdict`. Nothing re-renders the pair set or the win rate.
- **Capability reduction is structural.** The judge holds a one-method interface over `Router.CompleteFinal`, which omits tool definitions by request shape. There is nothing for it to call the REST-only unblinded pair view with. `TestEvalTools_JudgeSurfaceCannotUnblindAPair` is untouched; the internal equivalent asserts every judge request goes out with zero tools and no identity fields.
- **`judge_ident` is `judge_model`, not `operator`** - its verdicts count in the win rate and flip items to judged, and `Judgment.RubricVersions` still excludes operator calibration marks.

<details>
<summary>Config surface</summary>

```toml
[eval]
judge_model = "claude-sonnet-5"       # absent = no internal judge; MCP path unchanged
judge_provider = "anthropic"          # [[llm.providers]] name; empty = the agent's own
judge_max_cost_per_run = 0.50         # defaults to max_cost_per_run
```

`judge_provider` without `judge_model` fails validation - it would otherwise select nothing silently. An unknown provider name is rejected at `Start` via `Router.HasProvider`, so it fails the pass rather than every item.
</details>

<details>
<summary>Cost</summary>

Read from `CostTracker.SessionCost("eval:judge:{run}")`, never the provider-reported figure, and measured from a per-pass baseline - the key is stable, so a second pass would otherwise open already over its cap. Checked before each dispatch like the runner's cap, so an in-flight item always finishes. Accumulated on a new `eval_runs.judge_cost` column: judging is a separate budget spent by a separate decision, and folding it into `cost_spent` would make a judged run read as one that blew its cap. `Judge.StopAll` hangs off `OnPanic` beside `Runner.StopAll`.
</details>

<details>
<summary>Surfaces</summary>

`POST /api/v1/eval/runs/{id}/judge` (`eval:write`, background, terminal runs only; unset `judge_model` is a 503 naming the key). `GET /eval/config` reports the model, cap and rubric version, read from the judge's resolved config so an unwired judge is never advertised. `GET /eval/runs/{id}` gains `judging`. The Evals page's judgment-pending block gains a "Judge on the server" button beside the `/judge-eval` instruction - absent, not disabled, when no model is set. OpenAPI regenerated.
</details>

<details>
<summary>Tests</summary>

Go: blinding canary over the prompt itself, zero-tool-definitions, model override leaves the agent router alone, cost recorded apart from sample spend, cap stops dispatch with partial work kept, second pass gets a fresh budget, unreadable reply and provider error leave the item pending, not-configured, non-terminal, unknown provider, reply parsing. `TestRubricVersion_MatchesTheSkillFile` pins the Go rubric constant to `.claude/skills/judge-eval/SKILL.md`. Integration: run → internal judge → upgrade verdict against the mock LLM, with the judge actually reading the blinded payload so both presentation orders agree. Vitest for all three UI states. `just hook` green; no coverage threshold touched.
</details>

**Deliberately left out:**
- No retry on an unreadable judge reply - the item stays pending for a later pass, internal or from Claude Code.
- No WebSocket progress frame for a judging pass; the page polls the run detail's `judging` flag, the same "frames are droppable, the row is authoritative" stance the runner takes.
- `judge_model`/`judge_provider` are global, not per-agent. Nothing yet wants a per-agent judge.
- Threshold and judge config stay TOML + reload, per §11.7.
- The first real payload run against a live provider is still outstanding - this is covered in-process only.

https://claude.ai/code/session_01B5s2eiuUcCyhFGYxLjETsK


internal/eval/judge.go
```diff
@@ -0,0 +1,467 @@
...
+	if err != nil {
+		j.itemFailed(st, item, "building judge prompt", err)
+		return
+	}
+	wire := make([]llm.Message, 0, len(msgs))
+	for _, m := range msgs {
+		wire = append(wire, llm.Message{Role: m.Role, Content: m.Content})
+	}
+
+	resp, err := client.CompleteFinal(ctx, convID, wire)
```
Temikus: **[correctness — high] The router's session-cost guards act on the raw cumulative total of the stable `eval:judge:{run}` key, defeating the per-pass baseline.**

`CompleteFinal` → `completeInternal` runs two guards keyed on `costTracker.sessionCosts[sessionID]`, which never resets for this key:

1. `applyCostLimitFallback`: with a `[llm.fallbacks]` `trigger = "cost_limit"` rule configured, once cumulative judge spend on the run crosses the soft/hard session limit, every subsequent judge call is **silently rerouted to the fallback model/provider** — mid-pass and on every later pass — while the verdicts are still stamped `judge_ident=judge_model` / `RubricVersion`, and `JudgePass.Model` still reports the configured model. The "one derivation of the win rate" claim quietly becomes "two judges under one name".
2. `ExceedsHardLimit`: with a `[costs]` session hard limit (default or per-provider) below cumulative judge spend, every call for this run fails with `ErrHardLimitExceeded` — permanently, across passes — so the run becomes unjudgeable internally even though `passState.baseline` gives each pass a fresh `judge_max_cost_per_run` budget. `TestJudge_SecondPassIsNotBornOverBudget` passes only because the tests configure no `[costs]` limits.

The runner's samples don't hit this because each sample gets a fresh `eval:{run}:{task}:{k}:{variant}` key; the judge's key deliberately accumulates. Consider either a per-pass session key (e.g. `eval:judge:{run}:{pass}` — summing passes into `judge_cost` already works since `AddJudgeCost` accumulates) or exempting the judge session from the router's limit machinery the way its own cap is baseline-relative.

---

internal/eval/judge.go
```diff
@@ -0,0 +1,467 @@
...
+	j.mu.Lock()
+	defer j.mu.Unlock()
+	_, ok := j.active[runID]
+	return ok
+}
+
+// JudgeConvID is the cost tracker's session key for a run's judging spend,
+// distinct from every eval:{run}:{task}:{k}:{variant} sample key so judge cost
+// can never be mistaken for a sample's.
+func JudgeConvID(runID int64) string { return fmt.Sprintf("eval:judge:%d", runID) }
```
Temikus: **[correctness — medium] Judge spend is attributed to an agent literally named `"eval"`, not a `#`-marked pseudo-identity — nothing calls `RegisterSessionAgent` for this key.**

Eval sample sessions get `RegisterSessionAgent(convID, "{base}#eval:{variant}")` from `Engine.DryRun` (engine.go:1726), and `.claude/rules/eval-dryrun.md` leans on that: "The cost tracker also registers the pseudo-agent, so preview spend never lands in a real agent's totals" — `#` is the safety because `ValidResourceName` rejects it. The judge bypasses the engine, so `agentForSession("eval:judge:{run}")` falls through to the prefix parse and returns `"eval"`.

Consequences:
- `AgentCosts()` (the Costs dashboard) shows judge spend under a phantom agent named `eval`, unlike every other eval spend.
- `"eval"` is a perfectly valid agent name. An operator who has an agent called `eval` gets the judge's spend merged into that agent's totals, and — worse — that agent's `[costs]` per-agent limit overrides then govern the judge's calls via `limitsForSession`.

Fix: at pass start, register the pseudo-identity on the tracker, e.g. `router.CostTracker().RegisterSessionAgent(JudgeConvID(run.ID), run.BaseAgent+"#"+string(agent.ExecEval))` — the same identity `emitLifecycle` already writes to the audit log.

---

cmd/denkeeper/main.go
```diff
@@ -1499,6 +1499,27 @@ func buildEvalRunner(cfg *config.Config, store *eval.Store, dispatcher *agent.Di
 	}, logger)
 }
 
+// buildEvalJudge wires the internal judge over the same live engines the
+// runner uses, for its router: the judge overlay is a WithModel/WithProvider
+// clone, so judging bills to the agent's own cost tracker and honours its
+// pricing and fallback rules. It is unavailable — not absent — when
+// [eval] judge_model is unset, so the endpoint can say why.
+func buildEvalJudge(cfg *config.Config, store *eval.Store, dispatcher *agent.Dispatcher, auditor audit.Emitter, logger *slog.Logger) *eval.Judge {
```
Temikus, line 1506: **[correctness/docs — medium] Config reload never reaches the judge, but the PR says "judge config stay[s] TOML + reload".**

`buildEvalJudge` snapshots `cfg.Eval.Judge*` into `JudgeConfig` at startup, and `buildReloadFunc` re-applies engine knobs only — it never rewires the judge (or the runner, but the runner makes no reload claim in this PR). So after `POST /server/reload`:

- setting `judge_model` in TOML still leaves `POST /eval/runs/{id}/judge` returning 503 until a restart;
- changing `judge_model`/`judge_max_cost_per_run` keeps judging on the old model/cap;
- `GET /eval/config` becomes internally inconsistent: thresholds are read live from `s.deps.Config.Eval` (updated in place by reload) while the judge fields come from the stale snapshot.

Either re-resolve the judge config on reload (e.g. have `Judge` read from a getter/the shared `*config.Config` under a lock, or rebuild its `JudgeConfig` in `buildReloadFunc`), or drop the reload claim from the PR text and document restart-required — the silent no-op is the worst of the three options.

---

internal/api/eval.go
```diff
@@ -1031,3 +1035,70 @@ func (s *Server) handleEvalSuggest(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	id, ok := evalRunID(w, r)
+	if !ok {
+		return
+	}
+	var in evalJudgeInput
+	// An empty body is the ordinary case — judge everything outstanding — so it
+	// must not be an error.
+	if r.ContentLength > 0 {
```
Temikus: **[correctness/reuse — low] `r.ContentLength > 0` silently drops the body for chunked requests, and re-implements `decodeEvalBody`.**

`ContentLength` is `-1` for a chunked/unknown-length request (streaming clients, some proxies), so a real JSON body carrying `sample_n`/`limit` is silently ignored and the pass judges the whole queue — no error, wrong scope. The existing helper a few hundred lines up already handles exactly this case ("An empty body is accepted as an empty struct" via the EOF check):

```go
var in evalJudgeInput
if !decodeEvalBody(w, r, &in) {
    return
}
```

Same behavior for empty bodies, correct behavior for chunked ones, and one less bespoke decode path in the file.

---

web/src/components/EvalResults.svelte
```diff
@@ -161,6 +173,51 @@
...
+        detail = await api.evalRun(run.id)
+      } catch {
+        // A failed poll is not a failed pass: keep waiting.
+        continue
+      }
+      if (!detail?.judging) break
+    }
+    if (destroyed) return
+    await load()
+    serverJudge = { state: 'done', message: 'Judging finished.' }
```
Temikus: **[correctness (UI) — low] Exhausting the poll budget reports "Judging finished." while the server still says `judging: true`.**

`pollJudging` loops `JUDGE_POLL_LIMIT`(150) × 2 s ≈ 5 minutes. A full queue is "hundreds of items" per the PR (50 tasks × k=3 × two orders = 300), so at `max_concurrent = 2` and a few seconds per completion, a perfectly healthy pass routinely outlives the budget. When the loop falls through without ever seeing `judging: false`, the code still sets `state: 'done', message: 'Judging finished.'` and re-enables the button — a false completion, and a second click then surfaces a 409 ("already being judged") styled as a failure.

Track whether the loop exited via `break` vs. exhaustion, and on exhaustion say something honest ("Still judging — numbers update as pairs land") while leaving the button disabled, or keep polling at a slower cadence.

---

web/src/components/EvalResults.svelte
```diff
@@ -16,6 +16,12 @@
     // True when the run covered a subset of the set, i.e. a Quick check. A
     // clean Quick check earns the escalation CTA.
     quick = false,
+    // The internal judge's model, from GET /eval/config. Empty means [eval]
+    // judge_model is unset and Claude Code over MCP is the only judge path, so
+    // the server-side affordance is absent rather than disabled — there is
+    // nothing the operator could click to fix it from here.
+    judgeModel = '',
```
Temikus, line 23: **[correctness (UI) — low] The `judging` flag this PR added to `GET /eval/runs/{id}` is never consumed, so a page opened mid-pass shows an idle button.**

The Go comment on `evalRunDetail.Judging` says it exists "so a page can say why pairs_judged is climbing on its own" — but `EvalResults` only learns about a pass it started itself (`serverJudge` state). Reload the page (or open it from another tab) while a pass is running and: the "Judge on the server" button renders enabled, clicking it gets a 409 rendered as a red failure, and the climbing judged counts have no explanation.

On mount (or when the pending block renders), fetch `api.evalRun(run.id)` once; if `judging` is true, seed `serverJudge = { state: 'running', message: … }` and start `pollJudging()` so a reload rejoins the pass instead of colliding with it.

---

internal/eval/judge.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+	if err := j.store.AddJudgeCost(ctx, run.ID, delta); err != nil {
+		j.logger.Warn("recording eval judge cost failed", "run", run.ID, "error", err)
+	}
+}
+
+// sessionCost reads the pass's true spend from the agent router's cost
+// tracker, for the same reason a sample does: provider-reported cost is only
+// filled in by OpenRouter, so a cap keyed on it would never trip elsewhere.
+func (j *Judge) sessionCost(baseAgent, convID string) float64 {
```
Temikus: **[correctness — low] `sessionCost` re-resolves the engine on every call, so an agent deleted (or renamed) mid-pass silently kills the cost cap and the `judge_cost` deltas.**

`clientFor` captures the router clone once at `Start`, so completions keep flowing and keep spending after `DELETE /agents/{name}` removes the base agent from the dispatcher — but `overCap`/`recordCost` then get `j.engines(baseAgent) → !ok → 0`: `spent = 0 − baseline` never reaches `MaxCost` (cap dead for the rest of the pass), and `delta = 0 − recorded ≤ 0` so no further spend lands on `eval_runs.judge_cost`.

Resolve the tracker once at `Start` next to the client (e.g. keep the `Engine`, or the `*llm.CostTracker`, on the pass state) so the readings come from the same place the spend is recorded — the runner holds its `Engine` for the duration of a run for the same reason.

---

internal/eval/judge.go
```diff
@@ -0,0 +1,467 @@
...
+	// base agent's own.
+	Provider string
+	// MaxCost caps one judging pass in USD.
+	MaxCost float64
+	// MaxConcurrent bounds items in flight, shared with nothing: the runner's
+	// semaphore guards sample dispatch, and a judging pass runs after a run has
+	// finished.
+	MaxConcurrent int
+}
+
```
Temikus, line 85: **[design/limits — low] Nothing bounds judging passes across runs: each pass gets its own `MaxConcurrent` semaphore, so total in-flight judge calls are `MaxConcurrent × active passes` (plus any run's samples).**

The comment says "a judging pass runs after a run has finished", but that only holds per run: judging three finished runs at once (or judging run A while run B is still executing samples) puts `3 × max_concurrent` judge completions plus the runner's samples in flight simultaneously. `max_concurrent` is documented for the runner as "eval samples running at once, **process-wide**"; an operator who set it to protect a rate-limited provider will be surprised that the judge multiplies it per pass.

A shared semaphore on the `Judge` (constructed once in `NewJudge` instead of per `run()`), or one shared with the runner, restores the process-wide meaning.

---

cmd/denkeeper/main.go
```diff
@@ -1499,6 +1499,27 @@ func buildEvalRunner(cfg *config.Config, store *eval.Store, dispatcher *agent.Di
...
+	source := func(name string) (eval.Engine, bool) {
+		e := dispatcher.Agent(name)
+		if e == nil {
+			return nil, false
+		}
```
Temikus, line 1532: **[reuse — nit] This `source` closure is a verbatim copy of the one in `buildEvalRunner` (line 1484).**

The harness change in this same PR extracts the shared `source` and hands it to both `NewRunner` and `NewJudge`; main.go should do the same — either a small `engineSource(dispatcher)` helper or building runner and judge in one function. The typed-nil guard comment on `eval.EngineSource` is load-bearing ("the nil check there must happen *before* the value is boxed"), so having two copies is two places for that subtlety to drift.

---

internal/api/server.go
```diff
@@ -291,6 +292,9 @@ func New(cfg config.APIConfig, deps Deps, logger *slog.Logger) *Server {
 	// Unblinded, and deliberately REST-only: the judge's MCP surface must not
 	// be able to look up which variant produced which response.
 	mux.HandleFunc("GET /api/v1/eval/runs/{id}/pairs", s.RequireScope("eval:read", s.handleEvalRunPairs))
+	// Judging spends real money, so it sits behind eval:write beside the run
+	// endpoints — not on the read side with the pair views.
+	mux.HandleFunc("POST /api/v1/eval/runs/{id}/judge", s.RequireScope("eval:write", s.handleJudgeEvalRun))
```
Temikus, line 306: **[docs — nit] `.claude/rules/rest-api.md`'s endpoint map was not updated with `POST eval/runs/{id}/judge`.**

The PR updates `eval-dryrun.md`, `llmstxt.go`, README and the website docs, but the eval row of the endpoint map in `.claude/rules/rest-api.md` (the file whose stated scope is "REST endpoint map, scopes, non-obvious endpoint semantics") still lists only `runs`, `stop`, `summary|samples|pairs`. Worth adding the judge endpoint with its `eval:write` scope and the "terminal runs only, 503 without judge_model, 409 while active" semantics — that's exactly the kind of non-obvious behavior the map exists to pin.

---

Temikus (commented):
Code review of PR #370 at **xhigh effort** — note: this was a **single-pass inline review done without the Agent tool** (no multi-agent fan-out, no independent subagent verify pass), so treat verification confidence accordingly. All 10 angles (line-by-line, removed-behavior, cross-file tracing, language pitfalls, wrapper correctness, reuse, simplification, efficiency, altitude, CLAUDE.md conventions) were worked through in one context.

The core review decision — does the internal judge stay incapable of unblinding its own queue, and does it share exactly one derivation with the MCP path — checks out: the `judgeLLM` one-method interface over `CompleteFinal` omits tools by request shape, the payload is `GetBlindedItem` serialized whole, and the queue/write path is `ListPending`/`RecordVerdict` verbatim. The findings below are about the *cost/limit plumbing* around that core, plus UI/doc gaps.

Findings, ranked most-severe first (details in the inline comments):

```json
[
  {"file": "internal/eval/judge.go", "line": 363, "summary": "Router session guards (cost_limit fallback, hard limit) act on the raw cumulative total of the stable eval:judge:{run} key, so they silently swap the judge model or permanently refuse calls despite the per-pass baseline", "failure_scenario": "With a [llm.fallbacks] cost_limit rule or a [costs] session hard limit configured, cumulative judge spend across passes crosses the session limit -> subsequent judge completions are silently rerouted to the fallback model (verdicts still stamped judge_model/rubric v1) or fail with ErrHardLimitExceeded on every later pass, making the run unjudgeable internally"},
  {"file": "internal/eval/judge.go", "line": 267, "summary": "Judge session is never registered with the cost tracker, so spend is attributed to an agent literally named 'eval' instead of a #-marked pseudo-identity", "failure_scenario": "agentForSession('eval:judge:5') prefix-parses to 'eval'; an operator with a real agent named eval (a valid resource name) gets judge spend merged into that agent's totals and that agent's [costs] overrides applied to judging — violating the eval-dryrun.md invariant that preview spend never lands in a real agent's totals"},
  {"file": "cmd/denkeeper/main.go", "line": 1507, "summary": "Config reload never reaches the judge though the PR claims judge config stays 'TOML + reload'; /eval/config mixes live thresholds with the stale judge snapshot", "failure_scenario": "Operator sets [eval] judge_model, POSTs /server/reload, sees 'config reloaded' — but POST /eval/runs/{id}/judge still 503s (and a changed judge_max_cost_per_run keeps the old cap) until a full restart, silently"},
  {"file": "internal/api/eval.go", "line": 1080, "summary": "r.ContentLength > 0 skips body decoding for chunked requests and re-implements the existing decodeEvalBody helper", "failure_scenario": "A chunked/unknown-length POST with {\"sample_n\":20} has ContentLength -1 -> body silently ignored -> the pass judges the entire queue instead of the calibration subset, with no error"},
  {"file": "web/src/components/EvalResults.svelte", "line": 218, "summary": "pollJudging reports 'Judging finished.' when the 150-poll (~5 min) budget is exhausted even though the server still reports judging", "failure_scenario": "A healthy 300-item pass at max_concurrent 2 outlives 5 minutes -> user is told judging finished while verdicts are still landing, button re-enables, second click surfaces a 409 styled as a failure"},
  {"file": "web/src/components/EvalResults.svelte", "line": 23, "summary": "The judging flag added to GET /eval/runs/{id} for the page's benefit is never consumed, so a reload mid-pass shows an idle button", "failure_scenario": "User starts a pass, reloads the page -> button enabled, judged counts climb unexplained, clicking the button returns 409 'already being judged' rendered as a red failure"},
  {"file": "internal/eval/judge.go", "line": 422, "summary": "sessionCost re-resolves the engine per call, so deleting the base agent mid-pass zeroes readings, disabling the cost cap and losing judge_cost deltas", "failure_scenario": "DELETE /agents/{name} during a pass -> completions continue on the captured router clone but overCap computes spent = 0 - baseline (never trips) and recordCost deltas go negative (never persisted) for the rest of the pass"},
  {"file": "internal/eval/judge.go", "line": 57, "summary": "Per-pass semaphores mean total in-flight judge calls are MaxConcurrent x active passes, breaking max_concurrent's documented process-wide meaning", "failure_scenario": "Operator sets max_concurrent=2 for a rate-limited provider, judges 3 finished runs at once -> 6 concurrent judge completions plus any running samples -> provider 429s"},
  {"file": "cmd/denkeeper/main.go", "line": 1512, "summary": "buildEvalJudge duplicates buildEvalRunner's EngineSource closure verbatim instead of sharing it as the harness in this same PR does", "failure_scenario": "Two copies of the typed-nil-guard subtlety the EngineSource doc calls load-bearing; a future edit to one copy drifts from the other"},
  {"file": "internal/api/server.go", "line": 297, "summary": ".claude/rules/rest-api.md's endpoint map was not updated with POST eval/runs/{id}/judge", "failure_scenario": "The invariants file whose scope is the REST endpoint map omits the new spend-triggering endpoint while every other doc surface (llms.txt, README, website) was updated — future sessions loading the rule get a stale map"}
]
```

No blocking blinding/unblinding defect found; findings 1–2 are the ones I'd want addressed before real money runs through this, since both undermine the cost story the PR body leads with.

https://claude.ai/code/session_01QVBqfwic6TtsBLTTBWeZcv

---

internal/eval/judge.go
Temikus: Fixed in 371992c. Two halves to this one, so two fixes:

1. The cost-tracker session key is now per-pass - `JudgeConvID(runID, pass)` gives `eval:judge:{run}:{pass}`, numbered off an `atomic.Int64` on the Judge. Each pass starts the router's guards from zero, so a long run no longer trips `cost_limit` fallback or the hard limit on its own history.
2. A silent swap is now loud: `judgeItem` checks `resp.Model` against `judge_model` via `servedByJudgeModel` and fails the item with `ErrJudgeModelSwapped` if the completion came back from something else. Prefix-tolerant in both directions so a dated or aliased id does not false-fail, and an empty `Model` (provider did not report one) is accepted.

Tests: `TestJudge_ModelSwappedUnderneathFailsTheItem`, `TestServedByJudgeModel_ToleratesAliasesAndSilence`, `TestJudge_EachPassSpendsUnderItsOwnSessionKey`.

internal/eval/judge.go
Temikus: Fixed in 371992c. `Start` now calls `tracker.RegisterSessionAgent(convID, JudgeAgentIdent(run.BaseAgent))` before the first completion, where `JudgeAgentIdent` yields `{agent}#eval:judge` - same `#`-marked shape as `{agent}#dryrun` and `{agent}#eval:{variant}`, so it cannot collide with a real agent name.

Test: `TestJudge_SpendIsAttributedToTheEvalPseudoAgent` asserts `AgentCosts()` carries `pamela#eval:judge` and never `eval` or bare `pamela`.

cmd/denkeeper/main.go
Temikus: Fixed in 371992c. Extracted `judgeConfigFrom(cfg)` and added `Judge.SetConfig`, and `buildReloadFunc` now calls it alongside `approvals.SetConfigRules`. The config snapshot is `sync.RWMutex`-guarded, so an in-flight pass keeps the config it started with.

One deliberate exception: `SetConfig` preserves `MaxConcurrent`, since the semaphore is sized at construction. Documented in `.claude/rules/eval-dryrun.md` and the config reference.

Test: `TestJudge_SetConfigTurnsTheJudgeOnWithoutARestart`.

internal/api/eval.go
Temikus: Fixed in 32c3e30. Now just `decodeEvalBody`, which already tolerates an empty body (`err.Error() != "EOF"`), with a comment on why the `ContentLength` guard was wrong.

Test: `TestJudgeEvalRun_ReadsAChunkedBody` builds the request off an `iotest.OneByteReader` so `ContentLength` is negative, and asserts the `sample_n` survives.

web/src/components/EvalResults.svelte
Temikus: Fixed in efae734. `pollJudging` tracks whether it actually saw the pass finish; on budget exhaustion it stays in `running` with "Still judging on the server. Reload the page for the latest count." instead of claiming completion.

web/src/components/EvalResults.svelte
Temikus: Fixed in efae734. Added `resumeJudging()` - reads `api.evalRun(run.id)` on mount and, when `detail.judging` is true, restores the running state and resumes polling.

Test: `picks a running pass back up after a page reload` (MSW returns `judging: true` on the first poll, then false) asserts both the status text and that the judge button is disabled.

internal/eval/judge.go
Temikus: Fixed in 371992c. `clientFor` became `sessionFor`, returning a `judgeSession` that captures both the `judgeLLM` and the `CostTracker` once at `Start`. Cost reads go through `session.cost(convID)`, so a mid-pass agent deletion or rename cannot quietly disable the cap.

internal/eval/judge.go
Temikus: Fixed in 371992c. The semaphore moved onto the `Judge` (`j.sem`, sized once at `NewJudge`), so `max_concurrent` bounds in-flight judge calls process-wide rather than per pass.

Test: `TestJudge_ConcurrencyBoundIsProcessWide` checks the capacity and that the channel identity is unchanged after a pass runs.

cmd/denkeeper/main.go
Temikus: Fixed in 371992c. Extracted `liveEngineSource(dispatcher)`; both `buildEvalRunner` and `buildEvalJudge` use it.

internal/api/server.go
Temikus: Fixed in d5b75fe. Added the `POST eval/runs/{id}/judge` row to the endpoint map with its scope (`eval:write`) and status codes (202/503/409), plus the note that it never widens the judge surface - the unblinded `/pairs` endpoint stays REST-only.

PR #371, opened 2026-08-26
eval: close the five Stage D carry-forward follow-ups

### LINKED ISSUE #357: Web UI: promote four duplicated CSS classes into shared.css, and document the confirm-pattern rule
Fallout from the `ui-reviewer` pass on the Evals results scorecard. None of this blocked that work, and all of it is codebase-wide rather than scoped to one component, so it wants its own change rather than riding along.

## 1. Four classes every component redefines locally

`web/src/shared.css` does not define these, so each component that needs one writes its own. Verified counts on `main` plus the D2 branch:

| Class | Local copies | Notes |
|---|---|---|
| `.sr-only` | 2 — `components/EvalResults.svelte:493`, `pages/Chat.svelte:1047` | The two have **drifted**: the Chat copy carries `padding: 0; margin: -1px; border: 0` and `clip: rect(0, 0, 0, 0)`; the newer one omits all four and uses `clip: rect(0 0 0 0)`. The Chat version is the correct one. |
| `.table-wrap` | 3 — `components/EvalResults.svelte:595`, `pages/Schedules.svelte:520`, `pages/Settings.svelte:407` | All three exist only to give a wide table its own `overflow-x: auto`. |
| `.inline-error` | 9 files | Mostly `color: var(--danger); font-size: 12px` with a different margin each time, but two genuinely diverge: `Approvals.svelte` uses monospace with `padding-left: 48px` and ellipsis truncation, and one file uses `0.8rem` rather than `12px`. |
| `.section-title` | 10 files | **Two different things share the name.** One shape is an 11px uppercase muted eyebrow (`SuggestCases.svelte:429`, `Evals.svelte:745`, `EvalResults.svelte:522`, `Channels.svelte:428`, `Collapsible.svelte:46`); the other is a 16px/600 heading (`Providers.svelte:634`, `ServerConfig.svelte:631`, `Overview.svelte:179`, `Costs.svelte:263`), with `Approvals.svelte:264` a 17px near-miss of the second. |

`SuggestCases.svelte:428` already carries a comment apologising for the situation ("Matches the page's own section headings rather than the shared sheet, which has neither"), which is a fair signal that the next component will make the same choice.

Suggested shape:

1. Promote `.sr-only`, `.table-wrap`, and `.inline-error` into `shared.css` under a Utilities heading, taking the most complete existing definition in each case, and delete the local copies. Where a local copy only differed by margin, keep the margin locally and drop the rest.
2. `.section-title` needs a decision before a move: the two shapes are different roles and should get different names (something like `.section-eyebrow` for the 11px form and `.section-title` for the 16px heading), then be promoted separately. This half is the larger job and could reasonably be split out.

## 2. Horizontal scroll regions are not keyboard reachable

Any `.table-wrap` that actually scrolls sideways can only be scrolled with a pointer — there is no focusable element inside the scroll container, so a keyboard-only user cannot reach the off-screen columns (WCAG 2.1.1 Keyboard). Affects the three files above; the six-column gate and category tables in `EvalResults.svelte` are the most likely to overflow in practice, at narrow widths.

Fix once in the shared class: `tabindex="0"` plus `role="region"` and an `aria-label` on the wrapper.

## 3. The overlay-vs-inline confirm rule is undocumented

The codebase has two confirm patterns and the split between them is sound, but it is only implied:

- **Overlay** (`.overlay` + `.confirm-modal` in `shared.css:77`) for irreversible actions — deleting a schedule (`Schedules.svelte:480`), stopping an eval run that has already spent money.
- **Inline** confirm rendered in place inside the card for reversible writes — `Providers.svelte:478`, `Agents.svelte:670`.

`shared.css:75` already heads the block "Confirmation modal (destructive actions)", which names half the rule but does not say what to do for the non-destructive case. That gap is how the Evals "Apply to agent" confirm was written as an overlay first (fixed since): it hid the gate table that was the whole justification for the change the operator was confirming.

Fix: extend the comment above `.overlay` to state both halves — overlay for irreversible actions only; inline confirm, in place, for reversible ones — and point at `Providers.svelte` as the reference for the second.

Temikus:
Closes #357, #358, #359, #360, #361 - the five carry-forward follow-ups filed from the Stage D review of the eval subsystem. Independent of each other, one commit each.

Two are behaviour changes worth a close read; the rest are additive or mechanical.

## Worth reviewing

**#358 - per-dimension winner letters resolved server-side.** `GET /eval/runs/{id}/pairs` resolved the overall winner but left `Dimensions` as the raw letters the judge saw, so the results view inferred the letter→variant mapping from each item's `winner_variant`. That inference is structurally unable to resolve an item every judge called a tie: no verdict on it names a variant. `PairVerdict` now carries `dimensions_variant` beside the raw map, built through the same `VariantFor` the winner goes through. The raw letters stay for anyone auditing what the judge was answering, and the client-side resolver is gone.

Blinding is unaffected: the assignment stays server-side, and `/pairs` is already the unblinded operator view that the judge's MCP tools cannot reach.

**#357 - shared CSS + the confirm-pattern rule.** `.sr-only`, `.table-wrap`, and `.inline-error` move into `shared.css`; locals keep only their own margin. `EvalResults`' `.table-wrapper` folds into `.table-wrap` - that was a fourth copy under a second name. Every wrapper gains `tabindex="0"`, `role="region"`, and an `aria-label`: a scroll container with nothing focusable inside it cannot be reached from the keyboard at all, so its off-screen columns were pointer-only (WCAG 2.1.1). The overlay-vs-inline confirm rule is now written down above `.overlay` and in CLAUDE.md's UI/UX standards.

## The rest

- **#359** - `?task_id=` on `/eval/runs/{id}/samples`, matching the filter already on `/pairs`. `Store.ListTaskSamples` sits beside `ListSamples` over one shared column list, and both endpoints' validation goes through one `evalTaskFilter`. The results view switched from a run-wide cache to a per-task fetch, so opening one row no longer pulls 222 traces.
- **#360** - a badge naming why a tool loop stopped (`max_rounds`, `repeated_calls`, `stop_requested`), so a wrap-up no longer reads as a completed answer. One change in `DryRunTranscript.svelte` serves both callers - schedule/skill previews and the eval diffs.
- **#361** - the provider-reported serving upstream on eval transcripts, suppressed when it merely repeats the model line.

## Deliberately left out

- **`.section-title`** (the second half of #357). Two different roles share the name across 10 files - an 11px uppercase eyebrow and a 16px heading - and they need a naming decision before a move. The issue flags this half as reasonably splittable.
- **`KV.svelte` / `Costs.svelte`** also define a `.table-wrapper`, but theirs adds a surface and a border. Different component, not another copy of this one.

<details>
<summary>Tests and verification</summary>

`just hook` green, uncached (`JUST_HOOK_FORCE=1`).

Go:
- `TestPairDetails_DimensionLettersResolveToVariants` - the two orders resolve the same variant from different letters.
- `TestPairDetails_DimensionsResolveOnAnAllTieItem` - the case the client resolver could not do.
- `TestPairDetails_UnreadableDimensionValueIsCarriedThrough`, `..._NoDimensionsLeavesResolvedMapEmpty`.
- `TestStore_ListTaskSamplesNarrowsWithoutLosingColumns` - narrowed read keeps every column and does not reach another run's rows.
- `TestEvalRuns_SamplesFilterByTask`, `..._SamplesFilterOnAnAbsentTaskIsEmpty`, `..._SamplesBadTaskIDIs400`.

Vitest: resolved dimension render, the all-tie item, the missing-map fallback; per-task fetch asserting one request for `task_id=11` across collapse-reopen; cut-short badge labelled, absent on a clean finish, only on the flaily side of a comparison, unknown slug not swallowed; upstream shown, suppressed when redundant, never on a preview transcript; every `.table-wrap` focusable and labelled.

OpenAPI spec regenerated for the two response-shape changes (#358, #359).

</details>

https://claude.ai/code/session_01UHNvVZHXmQsMsyeDSujWa3


web/src/components/EvalResults.svelte
```diff
@@ -208,15 +210,17 @@
       return
     }
     expandedTask = taskID
-    if (samples || samplesLoading) return
-    await loadSamples()
+    // A failed fetch left no entry, so reopening the row retries on its own.
+    samplesError = ''
+    if (samplesByTask[taskID]) return
```
Temikus: **[correctness — confirmed by reproduction] The per-task cache races with the still-global `samplesLoading`/`samplesError` scalars.**

The old guard `if (samples || samplesLoading) return` also covered the in-flight case; the new `if (samplesByTask[taskID]) return` only covers the cached case, so multiple `loadSamples` fetches can now be in flight at once while one boolean and one error string describe all of them. Concrete failures:

1. Open task A (fetch hangs), collapse it, open task B whose samples are already cached → B's row renders "Loading turns…" instead of its cached turns until A's fetch settles. I reproduced this deterministically in a vitest against this branch (`turns-loading` shown, `turn-11-0` absent).
2. Same shuffle with A's fetch failing → A's error banner (and a retry wired to B's task id) renders under B's row, replacing B's already-loaded turns.
3. Reopening a row while its own first fetch is still in flight issues a duplicate GET.

Fix: track in-flight/error per task (e.g. store a `{loading:true}` sentinel in `samplesByTask`, or key `samplesLoading`/`samplesError` by task id) and have the detail row read only its own task's state.

---

web/src/components/DryRunTranscript.svelte
```diff
@@ -32,6 +32,33 @@
...
+  // only for providers that route (OpenRouter). A candidate served by
+  // different upstreams between turns explains latency or quality variance
+  // within one variant, which is otherwise invisible. Suppressed when it just
+  // repeats the model line — "openrouter/kimi-k3 · OpenRouter" says nothing.
+  let upstream = $derived.by(() => {
+    const via = (transcript.upstream || '').trim()
+    if (!via) return ''
+    const model = (transcript.model || transcript.requested_model || '').toLowerCase()
+    const low = via.toLowerCase()
+    return model === low || model.startsWith(`${low}/`) ? '' : via
```
Temikus: **[correctness edge] The redundancy suppression can hide exactly the cross-turn upstream variance this chip exists to reveal.**

On OpenRouter the model id's vendor prefix names the model's maker, not the serving infrastructure — `anthropic/claude-3-opus` can be served by "Anthropic", "Google Vertex", or "Amazon Bedrock". With k=2, a turn served by "Anthropic" gets its chip suppressed (`model.startsWith("anthropic/")`) while the sibling turn served by "Google Vertex" shows `via Google Vertex`. The operator reading the two columns cannot tell whether the unlabeled turn had *no* upstream reported or a suppressed one — precisely the "served by different upstreams between turns" case the comment above names as the point of the feature.

Consider suppressing only on exact equality (`model === low`), or suppressing per-pair only when *every* shown turn's upstream is redundant.

---

web/src/shared.css
```diff
@@ -376,6 +387,39 @@
...
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
+
+/* Scroll container for a table wider than its column. Give it tabindex="0",
+   role="region" and an aria-label: a scroll region with nothing focusable
+   inside it cannot be reached from the keyboard at all (WCAG 2.1.1), and the
+   off-screen columns are then unreadable without a pointer. */
+.table-wrap {
```
Temikus, line 408: **[consistency / a11y] `Costs.svelte`, `KV.svelte`, and `Overview.svelte` keep the same WCAG 2.1.1 defect this rule fixes.**

All three have `overflow-x: auto` table scroll containers (`Costs.svelte:279`, `KV.svelte:248`, `Overview.svelte:197 .cost-table-wrapper`) with nothing focusable inside, so their off-screen columns stay pointer-only. The PR body's "different component, not another copy" note justifies not merging their *CSS*, but the keyboard fix (`tabindex="0"` + `role="region"` + `aria-label`) is independent of whether they share the class — a keyboard user on the Costs page still cannot reach the per-agent table's off-screen columns on a narrow window. Worth applying the three attributes there too (or filing the follow-up), since CLAUDE.md now states the rule.

---

web/src/components/__tests__/EvalResults.test.js
```diff
@@ -363,7 +455,35 @@ describe('EvalResults — per-task diffs', () => {
     await waitFor(() => expect(screen.getByTestId('pair-judgment-71')).toBeInTheDocument())
     const judgment = screen.getByTestId('pair-judgment-71')
     expect(judgment).toHaveTextContent('called it a tie')
-    expect(judgment.querySelector('.dimensions li')).toHaveTextContent('correctness: a')
+    expect(judgment.querySelector('.dimensions li')).toHaveTextContent('correctness: current')
+  })
+
+  // An older run judged before the resolved map shipped, or a value the server
```
Temikus: **[docs nit] "An older run judged before the resolved map shipped" cannot actually produce this fallback.**

`dimensions_variant` is computed at read time in `pairVerdicts`/`resolveDimensions` from the stored letters and the pair's assignment — it is not persisted per run — so runs judged before this change still come back fully resolved after an upgrade. The only real trigger for the raw-letter fallback is the second clause (a value the server carried through unresolved, or a missing variant name). The comment as written teaches a future maintainer the wrong persistence model; suggest dropping the first clause.

---

Temikus (commented):
Code review at extra-high effort. **Note on scope: this was a single-pass inline review done without the Agent tool (no multi-agent fan-out, no subagent verify pass)** — all 10 review angles were worked through in one context instead.

4 findings, most severe first (also left as inline comments):

1. **EvalResults.svelte:215 — confirmed:** the new per-task samples cache still shares one global `samplesLoading`/`samplesError`, and the old in-flight guard was dropped; concurrent fetches paint the wrong row's loading/error/empty state (reproduced with a deterministic vitest) and can double-fetch.
2. **DryRunTranscript.svelte:60:** upstream chip suppression via the model's vendor prefix hides one side of exactly the cross-turn upstream variance the chip exists to show, and makes "suppressed" indistinguishable from "not reported".
3. **shared.css/.table-wrap:** Costs, KV, and Overview keep the identical keyboard-unreachable scroll-container defect the PR fixes elsewhere; the "deliberately left out" rationale covers the CSS dedup but not the a11y attributes.
4. **EvalResults.test.js:461:** comment misstates when the raw-letter fallback can occur (`dimensions_variant` is computed at read time, so pre-upgrade runs still resolve).

Verified clean: server-side dimension resolution matches `VariantFor` semantics (letters normalized, ties preserved, unreadables carried through); blinding unaffected (`PairDetails` stays REST-only, no MCP consumer of `dimensions_variant`); `evalTaskFilter`/`ListTaskSamples` SQL and 400/404 semantics; swagger regeneration; the shared-CSS moves (every slimmed local rule has the global backing it, scoped margins intact); stop-reason slugs match the engine's. Go + vitest suites for the touched packages pass on the branch.

---

Temikus:
All four review findings are addressed on the branch (2553889 + bd38ebd):

- **Per-task samples race**: `samplesLoading`/`samplesError` are now keyed by task id with a per-task in-flight guard, so each detail row renders only its own state; tests cover the stalled-fetch-under-another-row cases and the duplicate-GET guard.
- **Upstream chip suppression**: the vendor-prefix `startsWith` heuristic is gone — the chip is suppressed only when the upstream exactly repeats the full model id (case-insensitive), so cross-turn serving variance on OpenRouter models stays visible. Tests pin the vendor-prefix-shows-chip and case-insensitive-repeat-suppressed cases.
- **Keyboard access for the remaining scroll containers**: `Costs.svelte` (both wrappers), `KV.svelte`, and `Overview.svelte` now carry `tabindex="0"`, `role="region"`, and aria-labels (styling untouched). No vitest suites exist for those pages, so no test additions there.
- **Raw-letter fallback comment**: corrected — the fallback triggers only on a value the server carried through unresolved, not on pre-upgrade runs.

`just hook` fully green on the pushed head.

---

PR #372, opened 2026-08-26
feat(eval): L1 turn trace capture plus a turn inspector

Temikus:
Stage E item 1 of `design/eval-subsystem.md` (§4.2, L1 trace capture): full turn traces plus a turn inspector in the dashboard.

**The decision to review first: `[eval] capture` defaults to `false`.** A trace holds everything the model saw, so live recording is an explicit operator choice. Eval samples are traced regardless (the judge reads the trace), and those turns never touch a live conversation. `TestParse_EvalCaptureDefaultsOff`, `TestTraceCapture_OffByDefault`, and `TestTraceCapture_NoSinkRecordsNothing` pin all three halves of that: config default, engine default, and "capture on but no sink is still no capture".

## What a trace holds

The built system prompt post-skill-injection, the history window as it went on the wire, per-round tool calls with arguments and results, the final response, timings, and usage. `ToolCallRecord.Arguments`/`.Result` are `db:"-"` and reach the database nowhere else, so this is their only durable home. The inspector is what answers "why did it do that" - the audit log has rounds and outcomes but never the payloads.

## Config surface

| Key | Default | Notes |
|---|---|---|
| `[eval] capture` | `false` | Record live turns |
| `[eval] max_trace_bytes` | `262144` | Oldest rounds dropped first, then oldest history; the trace records what went |
| `[eval] retention_days` | `30` | Its own knob, matching audit, not `[memory]`'s |

## Two design points worth a look

1. **The engine never persists a policy turn's trace.** It builds one (always, for dry runs and evals) and hands it back on `TurnResult.Trace`; `eval.Runner` writes it. Keeps `persistTurn`'s isolation boundary exactly where it was.
2. **`GET /traces` and `GET /traces/{id}` are `sessions:read`, not `eval:read`.** A trace is turn content, and the eval scopes belong to a judge that must never resolve a live prompt.

## Tested

`just hook` green. Go: store round-trip, filters, paging, retention prune, capture on/off/no-sink, policy-turn-never-persists, truncation ordering, config defaults and validation, both endpoints incl. the scope split and the 503. Vitest: 19 cases over the page's loading, both empty states, error, filters, expansion, truncation notice, and paging.

## Deliberately left out

- **No CSS promotion to `shared.css`.** The new page duplicates `.empty-state`, `.filters`, and `.load-more`, which is exactly what #357 already tracks. Doing it here would have mixed a cross-page refactor into a feature branch.
- **No record-replay** (Stage E item 6). This unblocks it, that is all.
- **No link from a trace to its session.** The conversation id renders as text; wiring it to the Sessions page is a follow-up, and eval conversation ids do not resolve there anyway.
- **The `conversation_id` filter has no UI.** It exists on the endpoint for that future link.


internal/eval/trace.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil, fmt.Errorf("listing turn traces: %w", err)
+	}
+	return rows, nil
+}
+
+// CountTraces returns how many traces are stored, so a list view can say
+// whether there is more behind the page it is showing.
+func (s *Store) CountTraces(ctx context.Context) (int, error) {
+	var n int
+	if err := s.db.GetContext(ctx, &n, `SELECT COUNT(*) FROM turn_traces`); err != nil {
```
Temikus: **[high] `CountTraces` ignores the listing filter, so `total` is wrong for any filtered list and the UI's "Load older turns" never terminates.**

`handleListTraces` returns `Total` from this unfiltered `COUNT(*)` while `Traces` is filtered. The page shows the load-more button whenever `traces.length < total`, so: with `source=eval` selected and, say, 10 eval traces among 500 total, after all 10 load the button stays visible forever — each press fetches `offset=50` against the 10-row filtered set, appends nothing, and `traces.length` (10) stays below `total` (500).

`CountTraces` needs to take the same `TraceFilter` (minus limit/offset) and share the WHERE-clause construction with `ListTraces`, the way the audit store shares `buildWhereClause` between `List` and `Stats`.

---

web/src/pages/Traces.svelte
```diff
@@ -0,0 +1,453 @@
...
+
+  async function loadMore() {
+    loadingMore = true
+    error = ''
+    // The offset only advances once the page it asked for is in hand: a failed
+    // read that moved it would silently skip a page on the next press.
+    const next = offset + limit
+    try {
+      const res = await api.traces({ limit, offset: next, source, agent: agentFilter })
+      traces = [...traces, ...(res.traces || [])]
```
Temikus: **[high] Offset paging under live inserts appends duplicate rows, and the keyed `{#each traces as row (row.id)}` breaks on duplicate keys.**

Rows are ordered `id DESC` with `LIMIT/OFFSET`. With capture on, any turn recorded between the first page and a "Load older turns" press shifts every offset: page 1 held ids 101..52, a new trace 102 lands, `offset=50` now returns 52..3 — id 52 is appended a second time. Svelte 5 throws `each_key_duplicate` for a keyed each with duplicate keys (dev builds crash the page; prod renders duplicate rows whose expand toggles fight over the same `id="trace-52"`).

The exact scenario this page exists for — watching live turns while paging — is the one that triggers it. Either dedup on append (`res.traces.filter(t => !seen.has(t.id))`) or page by cursor (`before_id` = smallest loaded id) instead of offset.

---

internal/api/traces.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		s.logger.Error("counting turn traces", "error", err)
+		writeJSON(w, http.StatusInternalServerError, map[string]string{"error": "internal server error"})
+		return
+	}
+
+	out := traceListResult{
+		Traces: rows,
+		Total:  total,
+		Limit:  filter.Limit,
```
Temikus, line 135: **[medium] The echoed `limit` is the requested one, not the effective one — a paging client using it skips rows.**

The store silently clamps `limit` to 200 and defaults 0 to 50, but the response echoes `filter.Limit` raw. `GET /traces?limit=500` returns 200 rows with `"limit": 500`; a client advancing `offset` by the echoed limit skips 300 traces per page. With no `limit` param the response says `"limit": 0` while 50 rows came back, contradicting the swagger text ("default 50, max 200").

Either echo the effective limit (have `ListTraces` return it, or apply the clamp/default in `traceFilterFrom` before the store call) or reject `limit > 200` with a 400 like `limit < 1` already gets.

---

web/src/pages/Traces.svelte
```diff
@@ -0,0 +1,453 @@
...
+
+  onMount(async () => {
+    await refresh()
+    try {
+      const res = await api.agents()
+      agents = (res.agents || []).map(a => a.name).filter(Boolean)
+    } catch { /* the agent chips are a convenience, not the page */ }
+  })
+
+  async function refresh() {
```
Temikus, line 67: **[medium] The list fetches have no stale-response guard, so interleaved `refresh`/`loadMore` responses corrupt the list.**

`toggle()` carefully guards its detail fetch (`if (expandedId !== id) return`), but the list fetches don't: press "Load older turns", then click a source chip before it resolves — `refresh()` replaces `traces` with the filtered first page, then the in-flight `loadMore` response (old filter, offset 50) appends its rows to the filtered list and advances `offset`. The result is a mixed-filter list with a wrong offset. Two rapid chip switches race the same way: whichever `refresh` resolves last wins, which can be the stale filter.

A monotonically increasing request token checked before applying each response (the same pattern as the detail guard) fixes all three interleavings.

---

web/src/pages/Traces.svelte
```diff
@@ -0,0 +1,453 @@
...
+  let expandedId = $state(null)
+  let detail = $state(null)
+  let detailError = $state('')
+  let loadingDetail = $state(false)
+
+  const sourceItems = [
+    { value: '',       label: 'All',      testid: 'source-all' },
+    { value: 'live',   label: 'Live',     testid: 'source-live' },
+    { value: 'eval',   label: 'Eval',     testid: 'source-eval' },
+    { value: 'dryrun', label: 'Dry run',  testid: 'source-dryrun' },
```
Temikus: **[medium] The "Dry run" source filter can never match anything — dry-run traces are never persisted.**

The only two `SaveTrace` call sites are `Engine.saveTurnTrace` (which refuses policy turns) and `Runner.saveSampleTrace` (eval samples only); the dry-run API handler builds a `dryRunTranscript` and drops `TurnResult.Trace`. So `source='dryrun'` rows cannot exist, yet this chip offers the filter, and the swagger `@Param source` plus `website/content/docs/reference/rest-api.md` advertise `dryrun` as a valid value. An operator who just ran a preview will select "Dry run", get a guaranteed-empty list, and read it as capture being broken.

Either drop the chip (and the doc claims) until something persists dry-run traces, or persist them.

---

internal/agent/trace.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	if payloadSize(p) > maxBytes {
+		// Last resort: the prompt, the system prompt and the response are all
+		// that is left. Clamp the two long ones rather than storing nothing.
+		tr.ClampedText = true
+		p.SystemPrompt = clampTraceText(p.SystemPrompt, maxBytes/2)
+		p.Response = clampTraceText(p.Response, maxBytes/4)
+		p.Prompt = clampTraceText(p.Prompt, maxBytes/8)
+	}
+	tr.Note = traceTruncationNote(tr, maxBytes)
```
Temikus, line 233: **[medium] `max_trace_bytes` is not actually a hard cap: the last-resort clamp never re-measures, and the truncation note is added after the final size check.**

Two gaps compound here:
1. `clampTraceText` cuts raw string bytes (maxBytes/2 + /4 + /8), but the size the cap is defined against is the *encoded* form. JSON escaping inflates content — control characters encode as `` (6 bytes per input byte), quotes/backslashes/newlines double — so a system prompt or response heavy in escapable bytes can leave the encoded payload at ~3x `maxBytes` after "truncation".
2. `tr.Note` and the `Truncation` struct are appended after the last `payloadSize` comparison, so even the plain path can land a few hundred bytes over.

The row still stores (`Bytes` reports the overrun honestly), but the documented invariant — "there is deliberately no unbounded value, since one pathological turn would otherwise fill the table" — is soft by up to a small multiple. A final re-check loop that halves the clamp budgets until `payloadSize(p) <= maxBytes` would make the cap real.

---

internal/agent/trace.go
```diff
@@ -0,0 +1,361 @@
...
+func truncateTracePayload(p TracePayload, maxBytes int) (TracePayload, bool) {
+	if maxBytes <= 0 {
+		maxBytes = DefaultMaxTraceBytes
+	}
+	if payloadSize(p) <= maxBytes {
+		return p, false
+	}
+
+	tr := TraceTruncation{}
+	for len(p.Rounds) > 0 && payloadSize(p) > maxBytes {
```
Temikus: **[medium/efficiency] The truncation loop re-marshals the entire payload once per dropped item, and a captured turn is marshaled three times end to end.**

`payloadSize(p)` marshals the whole payload; the two loops call it after every single dropped round/history message, so an oversized trace costs O(items × payload-size) JSON encodes. A pathological turn (the exact case truncation exists for — e.g. a 10MB tool result across 20 rounds) does ~20 multi-MB marshals inline in `chatWithApproval`, before the turn returns. On top of that, a normal captured turn is encoded three times: the fit check here, `t.Bytes = payloadSize(payload)` in `buildTurnTrace` (line 311), and `EncodePayload` in `SaveTrace`.

Cheaper: marshal once, and while over budget subtract each dropped element's own encoded size (an approximation is fine — one final verify marshal catches drift), then reuse those bytes for both `Bytes` and the stored blob.

---

internal/api/traces.go
```diff
@@ -0,0 +1,247 @@
...
+		Offset: filter.Offset,
+	}
+	if s.deps.Config != nil {
+		out.Capture = s.deps.Config.Eval.Capture
```
Temikus: **[low] Unsynchronized read of `s.deps.Config.Eval.*` races with hot reload's `*cfg = *newCfg`.**

`buildReloadFunc` overwrites the shared `config.Config` with a whole-struct copy on `POST /reload` while this handler reads `Eval.Capture`/`RetentionDays`/`TraceBytesCap()` from the same pointer with no synchronization — a data race under the Go memory model (torn read; `-race` flags it if a test ever exercises reload concurrently with a request). The engine side of this PR got it right (`traceMu` around `SetTraceCapture`/`traceSettings`); the handler side follows the existing unlocked `deps.Config` pattern other handlers use, so this extends a pre-existing problem rather than creating a new one — but these are three brand-new read sites on data whose whole point is to change at reload time.

---

web/src/pages/Traces.svelte
```diff
@@ -0,0 +1,453 @@
...
+              <div class="prompt-block">
+                <div class="block-label">Message that started the turn</div>
+                <pre class="block-body">{detail.prompt || '(empty)'}</pre>
+              </div>
+
+              <Collapsible title="System prompt as sent" id="trace-{row.id}-system" open={false}>
+                <pre class="block-body tall" data-testid="trace-system-prompt">{detail.system_prompt || '(empty)'}</pre>
+              </Collapsible>
+
+              <Collapsible title="History window as sent ({detail.history.length})" id="trace-{row.id}-history" open={false}>
```
Temikus, line 302: **[low] "History window as sent" always contains the current turn's message, duplicating the "Message that started the turn" block — and the "No preceding context" state is unreachable.**

A live turn's message is stored then read back as the last history entry, and a policy turn appends `currentTurn` explicitly, so `payload.History` always ends with the same text the page renders above as the trigger message (the trace test pins this: "its own message is the last history entry"). Consequences on this page: the count in "History window as sent (N)" overstates preceding context by one, the first turn of a conversation shows "(1)" containing a copy of the prompt block, and the `trace-history-empty` branch ("the model saw the system prompt and this turn") can only render for an empty-prompt turn.

Wire-accuracy is the right call for the payload; the page could either label the last entry as the current turn or render `history.slice(0, -1)` with the count adjusted.

---

cmd/denkeeper/main.go
```diff
@@ -2402,6 +2413,40 @@ func initAuditor(ctx context.Context, store *audit.SQLiteStore, cfg *config.Conf
 	return be, be.Close
 }
 
+// startTraceCleanupWorker runs periodic retention enforcement on captured turn
+// traces. They carry their own [eval] retention_days because they are the most
+// sensitive rows in the database; the cleanup *cadence* is shared with audit,
+// since one sweep interval per instance is plenty and a second knob would only
+// be another thing to get wrong.
+func startTraceCleanupWorker(ctx context.Context, store *eval.Store, retentionDays int, interval string, logger *slog.Logger) {
```
Temikus, line 2421: **[low/reuse] `startTraceCleanupWorker` is a line-for-line copy of `startAuditCleanupWorker` with only the store call and log strings changed.**

Both parse the interval, fall back to an hour, tick, compute the same cutoff, prune, and log. One generic worker taking a name and a `prune func(ctx, cutoff) (int, error)` closure would serve both and keep the cadence semantics from drifting. The copy has already drifted semantically in one spot: the `retentionDays <= 0 → unlimited retention` branch is reachable for audit but dead for traces (`applyEvalDefaults` turns 0 into 30 and validation rejects negatives), so the comment promises a knob the config surface doesn't offer.

---

internal/api/traces.go
```diff
@@ -0,0 +1,247 @@
...
+
+// Turn inspector endpoints (L1, design/eval-subsystem.md §4.2). They sit
+// behind sessions:read rather than eval:read: a trace is turn content — the
+// same class of data as GET /sessions/{id}/tool-calls — and the eval scopes
+// exist for a judge, which must never be able to read live prompts.
+
+// traceToolCall is one call in a trace, flattened out of its round so the
+// response drops straight into the dry-run transcript renderer the dashboard
+// already has. Same field names deliberately: one renderer, not two.
+type traceToolCall struct {
```
Temikus: **[low/reuse] `traceToolCall` re-declares `dryRunToolCall` field-for-field instead of reusing it.**

The comment says "Same field names deliberately: one renderer, not two" — but the way to guarantee that is one *struct*, not two kept in lockstep by hand. `dryRunToolCall` (dryrun.go) has the identical fields plus `Truncated bool `json:"truncated,omitempty"``, which is `omitempty` and harmless here (and the shared DryRunTranscript component actually reads `call.truncated`, so traces get the "Trimmed for display" affordance for free if a per-call cap ever lands). Reusing `dryRunToolCall` deletes this type and removes the drift risk the comment is worried about.

---

internal/api/traces.go
```diff
@@ -0,0 +1,247 @@
...
+		return
+	}
+	writeJSON(w, http.StatusOK, buildTraceDetail(row, payload))
+}
+
+// buildTraceDetail unpacks a stored trace into the response shape.
+func buildTraceDetail(row *eval.TraceRow, payload agent.TracePayload) traceDetail {
+	// The payload column has already been decoded; sending it twice would
+	// double the response for nothing.
+	row.Payload = ""
```
Temikus: **[low/simplification] `row.Payload = ""` is dead code with a misleading comment.**

`TraceRow.Payload` is tagged `json:"-"`, so it is never marshaled regardless — the payload could not be "sent twice" through the embedded struct. The assignment and the comment above it can go; if the intent was defense against a future tag change, a test asserting the detail response has no `payload` key would say so more durably.

---

web/src/pages/Traces.svelte
```diff
@@ -0,0 +1,453 @@
...
+      if (expandedId === id) loadingDetail = false
+    }
+  }
+
+  function fmtCost(v) {
+    return `$${(v || 0).toFixed(4)}`
+  }
+
+  function fmtDuration(ms) {
+    if (!ms) return '—'
```
Temikus: **[low] Falsy-zero: `if (!ms) return '—'` renders a real 0 ms latency as "—".**

`latency_ms: 0` (a sub-millisecond mock/local-provider turn, or a legacy row) displays as missing data rather than "0 ms". The shared `DryRunTranscript.duration()` gets this right with `ms == null`; matching that here (`if (ms == null) return '—'`) keeps the two duration formats consistent.

---

Temikus (commented):
Code review of #372 at extra-high effort (recall-oriented). **Note on process: this was a single-pass inline review — the Agent tool was unavailable in this session, so the usual multi-agent fan-out and independent verify pass did not run.** Every finding was re-checked against the diff and enclosing code by the same reviewer before posting.

13 findings, most-severe first:

1. `CountTraces` ignores the list filter → wrong `total`, and the UI's "Load older turns" never terminates under a filter (internal/eval/trace.go:192)
2. Offset paging under live inserts appends duplicate rows → keyed-each duplicate keys (web/src/pages/Traces.svelte:90)
3. Echoed `limit` is the requested value, not the effective (clamped/defaulted) one → API pagers skip rows (internal/api/traces.go:117)
4. List fetches lack the stale-response guard the detail fetch has → interleaved refresh/loadMore corrupt the list (Traces.svelte:62)
5. "Dry run" source filter (and the docs' `source=dryrun`) can never match — dry-run traces are never persisted (Traces.svelte:46)
6. `max_trace_bytes` is soft: raw-byte clamp vs encoded size + note added after the last size check → payload can exceed the cap severalfold (internal/agent/trace.go:207-212)
7. Truncation re-marshals the whole payload per dropped item; a captured turn is marshaled 3x on the live path (trace.go:196, 311)
8. New unsynchronized `deps.Config.Eval.*` reads race hot reload's `*cfg = *newCfg` (internal/api/traces.go:121; pre-existing pattern, three new sites)
9. History window always contains the current turn message, duplicating the Prompt block; the "No preceding context" state is unreachable (Traces.svelte:282)
10. `startTraceCleanupWorker` duplicates `startAuditCleanupWorker` verbatim, incl. an unreachable "unlimited retention" branch (cmd/denkeeper/main.go:2421)
11. `traceToolCall` re-declares `dryRunToolCall` field-for-field (internal/api/traces.go:21)
12. `row.Payload = ""` is dead (field is `json:"-"`) with a misleading comment (traces.go:212)
13. Falsy-zero in `fmtDuration`: real 0 ms renders as "—" (Traces.svelte:144)

Checked and found sound: capture off-by-default wiring (config default, engine default, no-sink), the typed-nil sink guard, the policy-turn persistence boundary (engine never writes, runner does), trace-before-persist ordering with raw `responseText` (reply-guard substitution only touches `out.response`), `saveTurnTrace`'s cancelled-context fallback, the sessions:read scope split, schema/migration approach, retention prune, rune-safe clamping, XSS-safety of the new page, and the swagger/docs/CLAUDE.md updates.

---

Temikus:
Thanks for the review - addressed 12 of the 13 findings in four commits (`40e520f`, `5ca6b96`, `781ed7e`, `678de8d`). `just hook` is green.

**Correctness**

1. **Filtered `total`** - `CountTraces` now takes the same `TraceFilter` as `ListTraces` and both share one predicate builder; paging fields stay out of the count. "Load older turns" terminates under a filter.
2. **Shifted pages** - `loadMore` drops rows already on screen before appending, so a turn recorded between two reads can no longer produce a duplicate key.
3. **Effective limit** - `eval.BoundTraceLimit` is exported and resolves the limit before it is used and echoed.
4. **Stale list reads** - both list fetches carry the sequence guard the detail fetch already had.
5. **Dry-run chip** - removed, and the docs/swagger now say `live`/`eval` with a note that a preview's trace is never stored.
6. **Hard byte cap** - the truncation note is attached before the final measurement (it rides inside the payload), and the last-resort clamp halves the text budget until the *encoded* payload fits rather than trusting one raw-byte guess. When the cap is smaller than the note itself the note is clamped too. New test asserts the encoded size stays under 4096/1024/256 byte caps with escape-heavy text.
9. **Duplicated prompt** - `traceHistory` drops the turn's own message, so the history block is preceding context only and the "no preceding context" state is reachable. New test covers a first turn (no history) and a second turn (the earlier exchange, not its own message).
13. **Zero duration** - `fmtDuration` checks the type, not truthiness.

**Efficiency and duplication**

7. One marshal establishes the payload size and the drop loops subtract what they remove; `buildTurnTrace` keeps the blob it measured, so the store no longer re-marshals it. Live path is down from three marshals to one.
10. `startTraceCleanupWorker` and `startAuditCleanupWorker` both call one `startRetentionWorker` that takes the prune as a function.
11. `traceToolCall` deleted; the detail response uses `dryRunToolCall` directly.
12. The dead `row.Payload = ""` and its misleading comment are gone.

**Not addressed**

8. **Unsynchronised `deps.Config` reads.** Real, but it is the pattern every handler in `internal/api` already follows, and hot reload's `*cfg = *newCfg` races all of them equally. Fixing it only at the three new sites would leave the codebase inconsistent without closing the race. Worth its own change that puts the config behind an accessor across the package - happy to file it if you want it tracked.

Invariants in `.claude/rules/eval-dryrun.md` updated with the filtered count, the hard cap, the preceding-context history rule, and the dry-run-never-stored fact.

Temikus:
Follow-up in f797441 on top of 678de8d, after re-verifying all 13 findings against the branch — the four fix commits check out as described (filtered count shares one predicate builder with the listing; append dedupe; effective limit via the exported bound; hard encoded-byte cap with the note inside the measurement; single-marshal truncation; preceding-context-only history; shared retention worker; `dryRunToolCall` reuse; dead code gone). Three gaps closed:

- **Finding 8 (config-read race) is now fixed rather than deferred**: the trace handlers read an `atomic.Pointer` snapshot seeded at construction and refreshed in the reload handler after a successful reload, so only the three new read sites moved and no package-wide refactor was needed. `TestTraces_ListDoesNotRaceConfigReload` fails under `-race` against the old read path and passes now; a second test pins that a reload's new capture settings are actually served.
- The stale-list-read guard from 781e7e1 had shipped untested — added the vitest that drops a late older-page response after a filter switch.
- The now-unreferenced `TraceSourceDryRun` constant is gone (a preview's trace value exists in memory only).

`just hook` fully green on the pushed head.

---

Temikus:
Filed the deferred finding 8 as #373 - config accessor plus a pointer-swap reload, with an audit of the other holders of that pointer.

PR #416, opened 2026-09-16
feat(agent): gate advertised tool definitions by skills' requires.tools

Temikus:
Once some active skill declares `requires.tools`, the turn advertises only the union of those declarations. Nothing declared, everything advertised - the inverse of the inclusion half, because a skill that declares nothing has said nothing about the turn's needs, and narrowing on silence would strip the tool surface off ordinary chat.

Guards:

- The union is read off the post-`filterUnsatisfiedSkills` set: a declared tool that does not exist has already deactivated its skill, so it contributes nothing rather than emptying the payload. One availability check, not two.
- Servers marked `MarkAlwaysAdvertised` are never gated: the per-agent config-MCP server and the reviewer's, so the agent can always patch its own bad declaration. KV rides along on that server.

The filter lives in the router, so the fallback retry inherits the narrowed payload and `CompleteFinal` stays tools-free by construction. A nil filter takes the pre-existing branch.

Divergence from plan D1, which gates on `drivingSkill` and exempts ambient: the sprint item asked for the union over all active skills, so an ambient declaration narrows ordinary turns too. WDYT?

`just hook` is green except `openapi-check`, which fails on main too (swag vs the Go 1.27 stdlib). No `internal/api` files here.

https://claude.ai/code/session_01C131ZKgwTJuWmkBzhAje2v


Temikus:
@copilot resolve the merge conflicts on this branch.

PR #417, opened 2026-09-16
feat(tools): per-server guidance injected into the system prompt

Temikus:
Operators can now put a server's argument conventions in `[tools.*] guidance` instead of pasting them into every skill body. The text lands in each agent's system prompt as a `## Tool Server Notes` section, one subsection per guided server, listing that server's advertised tool names and then the operator's text verbatim.

This is audit finding C1, carried across three agent-ops windows: fabricated `workspaceId`, a KV cache key passed where a project ID belongs, invalid `!section:` filter grammar. Guidance is prompt text only, never a validator.

**Where it lands**: right after the KV block, before `## Session Context` — guidance changes only on operator edit, so it belongs early in the prompt-cache prefix.

**Scoping**: a server contributes only when it declares guidance *and* currently advertises at least one enabled tool. Disabled, `config_error`, crashed and fully-`disabled_tools` servers fall out of that rule without being special-cased, as do in-process servers. Rendering is sorted and deterministic; with nothing guided the prompt is byte-identical to before.

**Caps**: 2000 bytes per server — a 400 on the REST/UI path, truncated on a rune boundary with a warning at load (never via `ToolWarnings`, which would disable the server). Plus an 8000-byte aggregate cap and a 40-name cap on the per-server tool list.

Surfaced read-only in `GET /tools`, `GET /tools/{name}` and the dashboard Tools page; `PUT` preserves it when the body omits the key, so a UI edit cannot erase it.

The supervisor's review prompt gets the owning server's guidance too (verbatim, per-tool lookup) — both `workspaceId` fabrications in the audit were supervisor-approved, so leaving the reviewer blind had a demonstrated cost.

Out of scope: per-tool or agent-authored guidance.

https://claude.ai/code/session_01C131ZKgwTJuWmkBzhAje2v


Temikus:
@copilot resolve the merge conflicts on this branch.

